PR #315, opened 2014-12-29
Fix how icebox maps are generated

daniviga:
Fixes:
- Missing basemaps / basemaps have wrong names
- Sometimes a geoserver error is triggered
- Error during parallel maps creation
- Wrong map bbox and zoom

Probably fixes (more work needed):
- Missing thumbnails for icebox maps

TODO:
- ~~Enable only one layer on the map by default~~
- More testing


nastasi-oq:
LGTM

PR #322, opened 2015-01-05
Revert GEOSERVER_URL to absolute URL, relative gives a 404 in GXP

daniviga:
Revert changes made in #217 because this configuration gives an 404 when a non superuser tries to add a layer to a map:

screenshot from 2015-01-05 16 01 53

We need to take this revert into account when we change protocol and or hostname. A change in the platform urls will break all the maps. Probably it would be better to solve the issue in other ways.

### DIFF
--- a/openquakeplatform/openquakeplatform/local_settings.py.template
+++ b/openquakeplatform/openquakeplatform/local_settings.py.template
@@ -126,7 +126,7 @@ TILESTREAM_URL = '//tilestream.openquake.org'
 
 HELP_URL = 'http://www.globalquakemodel.org/openquake/support/documentation/platform'
 
-# The GEOSERVER_URL must be an absoulute URL to let GXP working correctly.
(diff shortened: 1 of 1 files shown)


daniviga:
There are several other problems related to the URL change (in geoserver, tomcat...), so this fix it's ok for the moment.

openquakeplatform/openquakeplatform/local_settings.py.template
```diff
@@ -137,7 +137,7 @@ OGC_SERVER = {
     'default': {
         'BACKEND': 'geonode.geoserver',
          # Do not change the internal channel
-        'LOCATION': 'http://localhost:8080' + GEOSERVER_URL,
+        'LOCATION': 'http://localhost:8080/geoserver/',
```
nastasi-oq: Please, add some comment to explain side effects of this change.

nastasi-oq:
LGTM

PR #323, opened 2015-01-05
ecd: wrong id passed to find an index, fixed.

nastasi-oq:
This bug produced "Internal server error" for location with this particular type of sub-section.


daniviga:
LGTM

PR #324, opened 2015-01-07
Add support for Google Analytics using local_settings.py

daniviga:
Code is enabled only when a `UA-*****` is provided through the `local_settings.py`.

The `GOOGLE_UA` in settings is not mandatory.

### DIFF
--- a/openquakeplatform/openquakeplatform/local_settings.py.template
+++ b/openquakeplatform/openquakeplatform/local_settings.py.template
@@ -383,3 +383,5 @@ ICEBOX_MAX_SIZE_MB = 10
 ICEBOX_SEND_EMAIL = False
 
 AVATAR_GRAVATAR_DEFAULT = 'http://ftp.openquake.org/oq-platform/oq-avatar-40.png'
+
+GOOGLE_UA = False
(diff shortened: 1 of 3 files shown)


openquakeplatform/openquakeplatform/local_settings.py.template
```diff
@@ -383,3 +383,5 @@ ICEBOX_MAX_SIZE_MB = 10
 ICEBOX_SEND_EMAIL = False
 
 AVATAR_GRAVATAR_DEFAULT = 'http://ftp.openquake.org/oq-platform/oq-avatar-40.png'
+
+GOOGLE_UA = False
```
nastasi-oq: Comment it instead of assign directly with the default value (see below)

openquakeplatform/openquakeplatform/utils.py
```diff
@@ -32,9 +31,12 @@ def oq_context_processor(request):
     context['is_gem_experimental'] = settings.GEM_EXPERIMENTAL
     context['TILESTREAM_URL'] = settings.TILESTREAM_URL
     context['HELP_URL'] = settings.HELP_URL
+    if hasattr(settings, 'GOOGLE_UA'):
+        context['GOOGLE_UA'] = settings.GOOGLE_UA
```
nastasi-oq: use single line assignement with default value if not defined using ```getattr``` function.

nastasi-oq:
LGTM

PR #326, opened 2015-01-09
Replace Leaflet 'latest-master' with latest stable version v0.7.3

daniviga:
Latest Leaflet master seems to have problems with ECD application causing high CPU usage and memory leaks in Chrome with certain configurations.

Needs testing on applications that use Leaflet.


daniviga:
Paul confirmed that Leaflet stable performs much better with ECD (no memory leaks). Waiting ok from @bwyss

bwyss:
LGTM

PR #327, opened 2015-01-09
override the leaflet contorl layer labels override

bwyss:
This PR resolves a small css conflict between the labels and the leaflet layer control labels.


daniviga:
LGTM

PR #329, opened 2015-01-11
vuln: more consistent list page, no 'More details' button when no curve ...

nastasi-oq:
...selected and fix wrong default for curve_id.

### DIFF
--- a/openquakeplatform/openquakeplatform/vulnerability/views.py
+++ b/openquakeplatform/openquakeplatform/vulnerability/views.py
@@ -54,8 +54,9 @@ def list_entries(request, **kwargs):
                   types_of_assessment[type_of_assessment].lower())
 
     try:
-        curve_id = int(request.GET.get("curve",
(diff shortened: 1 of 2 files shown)


openquakeplatform/openquakeplatform/vulnerability/views.py
```diff
@@ -54,8 +54,9 @@ def list_entries(request, **kwargs):
                   types_of_assessment[type_of_assessment].lower())
 
     try:
-        curve_id = int(request.GET.get("curve",
-                                 TA.FRAGILITY))
```
nastasi-oq: The second argument of ```get``` method is the default value to use when ```curve``` argument isn't set, the correct value is ```-1``` not ```TA.FRAGILITY```.

bwyss:
I am able to see 'Select a list item on the left' for just a brief moment, then the 'more details' button is shown before any function is selected

bwyss:
LGTM

PR #330, opened 2015-01-12
Rename ISC field 'smajaz' to 'smajax' as in ISC-GEM v1.05

daniviga:
The original data contained a wrong column name.


bwyss:
LGTM

PR #338, opened 2015-01-16
Make version number more consistent

daniviga:
Make version number more consistent. Avoid dupplication between setup.py and __init__.py


micheles:
LGTM

PR #343, opened 2015-01-16
Add a "Download" button in documents view

daniviga:
screenshot from 2015-01-16 18 17 14
 
Fixes a common user request.


pslh:
LGTM

PR #344, opened 2015-01-16
Update the Geodetic Strain map to version 2.1

daniviga:
fd214a337be51206d71f64756f2b561c

Using: http://tilestream.openquake.org/#!/map/geodetic-strain-v2-1


bwyss:
LGTM

PR #345, opened 2015-01-17
back-compatibility with Leaflet stable version

nastasi-oq:
I search whole code to find the same error pattern without succes. Seems appears just here.


daniviga:
LGTM

PR #351, opened 2015-01-20
Hrde app bug

bwyss:
Rename the strain layer, and remove the strain arrows layer. Please note that the new strain layer contains both strain and strain arrows, the strain arrows are only render at zoom level 5 and above.


nastasi-oq:
LGTM

PR #354, opened 2015-01-20
Modify the data formatter

bwyss:
In the case that a discrete Vulnerability function has y-axis labels less than 0.2 we need to modify the data formatter.


nastasi-oq:
LGTM

PR #359, opened 2015-01-22
When exporting socioeconomic data, enable to filter by a list of country iso codes

ptormene:
This addresses https://bugs.launchpad.net/oq-platform/+bug/1411611 and is related to https://bugs.launchpad.net/oq-svir/+bug/1411613
* Add API method to list countries and iso codes (useful to populate in QGIS a list of countries for which socioeconomic data are available)
* when exporting socioeconomic data, accept an additional optional parameter to refine the filtering by iso codes (I renamed a API method, but the old name is still available for compatibility with the old client versions, and the old workflow still works)

### DIFF
--- a/openquakeplatform/openquakeplatform/svir/views.py
+++ b/openquakeplatform/openquakeplatform/svir/views.py
@@ -141,8 +141,8 @@ def export_variables_info(request):
     # We don't need 'Tag' anymore, but we need to show other fields
     # We don't display update_periodicity and internal_consistency_metric
     response = HttpResponse(content_type='text/csv')
-    content_disp = 'attachment; filename="socioeconomic_indicators_export.csv"'
-    response['Content-Disposition'] = content_disp
+    response['Content-Disposition'] = \
+        'attachment; filename="socioeconomic_indicators_export.csv"'
     name_str = request.GET.get('name')
     keywords_str = request.GET.get('keywords')
     theme_str = request.GET.get('theme')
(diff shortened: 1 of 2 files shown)


openquakeplatform/openquakeplatform/svir/views.py
```diff
@@ -217,16 +217,45 @@ def export_variables_info(request):
...
 @sign_in_required
-def export_variables_data_by_ids(request):
+def export_countries_info(request):
+    """
+    Export a csv file containing iso codes and names of countries for which
+    socioeconomic data are available
+    """
+    response = HttpResponse(content_type='text/csv')
+    content_disp = 'attachment; filename="countries_info_export.csv"'
+    response['Content-Disposition'] = content_disp
```
nastasi-oq: Please, assign the string directly to the response item.

openquakeplatform/openquakeplatform/svir/views.py
```diff
@@ -235,6 +264,11 @@ def export_variables_data_by_ids(request):
...
         response = HttpResponse(msg, status="400")
         return response
+    country_iso_codes = request.GET.get('country_iso_codes')
+    country_iso_codes_list = []
+    if country_iso_codes:
+        country_iso_codes_list = [iso.strip()
+                                  for iso in country_iso_codes.split(',')]
     response = HttpResponse(content_type='text/csv')
     content_disp = 'attachment; filename="sv_data_by_variables_ids_export.csv"'
     response['Content-Disposition'] = content_disp
```
nastasi-oq: the same as above ?

nastasi-oq:
Except for those minors LGTM

PR #362, opened 2015-01-27
Remove support to SES in icebox since it's broken

daniviga:
Remove support to SES in icebox since this feature is currently broken and we don't visualize any output of type SES. This is the companion to https://github.com/gem/oq-engine/pull/1643


micheles:
LGTM

PR #364, opened 2015-01-27
Ecd fix scrollable

nastasi-oq:
Add active resize of list of events and list of locations and place "back" links in ecd toolbar only.


daniviga:
LGMT

PR #366, opened 2015-01-28
SVIR: add API to get the url needed to download a layer's TC211 metadata

ptormene:
Needed by https://github.com/gem/oq-svir-qgis/pull/11
(see also https://bugs.launchpad.net/oq-svir/+bug/1411598)


micheles:
LGTM

PR #370, opened 2015-02-16
Generate Icebox maps only when there are supported geo outputs

daniviga:
Also improve a bit the jobs list template.

### DIFF
--- a/openquakeplatform/openquakeplatform/icebox/models.py
+++ b/openquakeplatform/openquakeplatform/icebox/models.py
@@ -72,14 +72,14 @@ class Calculation(models.Model):
         # we create a database View per each output, then we publish the
         # db view on geoserver as a postgis vector layer. Finally, we
         # let geonode slurp the geoserver layer
-        if len(self.outputlayer_set.all()) > 0:
-            for olayer in self.outputlayer_set.all():
-                model = olayer.output_type
-                if model:
-                    layers.append(
-                        olayer.create_geonode_layer(
-                            olayer.geoserver_publish_view(
-                                olayer.create_view())))
+        for olayer in self.outputlayer_set.all():
+            model = olayer.output_type
+            if model:
+                layers.append(
+                    olayer.create_geonode_layer(
+                        olayer.geoserver_publish_view(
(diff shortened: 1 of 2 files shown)


openquakeplatform/openquakeplatform/icebox/models.py
```diff
@@ -72,14 +72,15 @@ def process_layers(self):
         # we create a database View per each output, then we publish the
         # db view on geoserver as a postgis vector layer. Finally, we
         # let geonode slurp the geoserver layer
-        for olayer in self.outputlayer_set.all():
```
ptormene: If `self.outputlayer_set.all()` returns `None`, nothing should happen inside the loop, so the additional check looks redundant. What does `self.outputlayer_set.all()` return, motivating that change?

daniviga: I need to skip `self.create_geonode_map(layers)`, which is outside the loop, when `len(self.outputlayer_set.all())` is less then 1.
There's no reason to try the loop is there are no layers, and at the same time I skip the map creation.

ptormene: I propose:
```
all_olayers = self.outputlayer_set.all()
for olayer in all_olayers:
    ...
if len(all_olayers) > 0:
    self.create_geonode_map(layers)
```

openquakeplatform/openquakeplatform/icebox/templates/icebox.html
```diff
@@ -97,7 +99,7 @@
             <% } %>
           <% } %>
 
-          <% if(calc.get('status') == 'complete' || calc.get('status') == 'failed' || calc.get('status') == 'queued') { %>
+          <% if(calc.get('status')) { %>
```
ptormene: So, if you have a status, whatever it is, you want the final href with the calc id to be there, otherwise you want the button to display the traceback in case the status it `failed`. If you don't have a status, you don't want to display the calc id. Am I understanding correctly?

ptormene:
LGTM

PR #372, opened 2015-02-17
update 'django-chained-selectbox' dependency version to 0.2

nastasi-oq:
fix wrong insert bug on vulnerability nested-inlines hierarchical dropdown menu


daniviga:
LGTM

PR #376, opened 2015-02-26
Icebox layer deletion

daniviga:
This PR will trigger layer and maps removal when an Icebox calculation is deleted. It will also DROP the database view associated with the icebox output.

We are missing a migration to add layers previously produced by Icebox in the "icebox" layers group. This will be part of a separate PR.

Thanks @ptormene for the peering.

### DIFF
--- a/openquakeplatform/openquakeplatform/icebox/models.py
+++ b/openquakeplatform/openquakeplatform/icebox/models.py
@@ -157,14 +157,15 @@ class Calculation(models.Model):
     def remove_calc(sender, instance, using, **_kwargs):
         if instance.map_id:
             for layer in instance.map.layer_set.all():
-                layer_obj = Layer.objects.get(typename=layer.name)
-                if layer_obj.store == 'icebox':
-                    cursor = connection.cursor()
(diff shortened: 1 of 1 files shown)


daniviga:
Migration: https://github.com/gem/oq-platform/pull/377

openquakeplatform/openquakeplatform/icebox/models.py
```diff
@@ -154,8 +155,16 @@ def create_geonode_map(self, layers):
...
 
     @staticmethod
-    def remove_map(sender, instance, using, **_kwargs):
+    def remove_calc(sender, instance, using, **_kwargs):
         if instance.map_id:
+            for layer in instance.map.layer_set.all():
+                if layer.group == 'icebox':
+                    layer_obj = Layer.objects.get(typename=layer.name)
+                    cursor = connection.cursor()
+                    view_name = layer.name.split(":")[1]
```
ptormene: We might add a comment with an example like geonode:layername

ptormene:
Apart from the minor note, LGTM

ptormene:
LGTM

PR #377, opened 2015-02-26
Add migration 002 for icebox layers

daniviga:
This migrations adds all the layers produced by an icebox calculation in the 'icebox' group.

We will need to add support for sql migrations in `migrate.sh`.


ptormene:
LGTM

daniviga:
Changing the layers `group` breaks GeoExplorer

daniviga:
Changing the layers group breaks GeoExplorer

PR #378, opened 2015-02-27
Irv refactor

bwyss:
This refactor changes the source of the data powering the app from UtfGrids to Geosever.
Currently the category (theme) and primary indicator charts are deactivated until more input is provided by the project owner.


openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,645 @@
...
-
-        for(var prop in pdTempPILevel) {
-            if(pdTempPILevel.hasOwnProperty(prop)) {
-                if(pdTempPILevel[prop] === foo) {
-                    return prop;
+    } else if (operator == 'Weighted multiplication') {
+        for (var v4 = 0; v4 < themeObj.length; v4++) {
+            var tempElementValue = 0;
+            var themeObjMunic = themeObj[v4].region;
+            // sum the themes
```
ptormene: Please change the comment

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,645 @@
...
+    var riWeight;
+    var iriOperator = projectDef.operator;
+    for (var ik = 0; ik < projectDef.children.length; ik++) {
+        if (projectDef.children[ik].name == 'RI') {
+            riWeight = projectDef.children[ik].weight;
+        } else if (projectDef.children[ik].name == 'SVI') {
+            sviWeight = projectDef.children[ik].weight;
+        }
+    }
+    if (iriOperator == "Average (ignore weights)") {
```
ptormene: I guess we might avoid using tempVal and do the calculations in one step

ptormene:
We might discuss about using the same name conventions, e.g.: in the QGIS plugin we call "zone" the geographic area of a single feature in the layer. I guess using words such as "district" might potentially cause confusion.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
-
-// Keep track of the utfGrid that has been selected last
-var selectedGrid;
+var layerAttributes;
+var sessionProjectDef = [];
+var selectedRegion;
+var sessionProjectDefStr;
+var projectLayerAttributes;
+var region = [];
+var districts = [];
```
ptormene: I guess this isn't used anymore. Please check

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
-        })) {
-            tempAalWeight = pdData.weight;
+    // Find the theme information
+    for (var m = 0; m < socialVulnIndex.length; m++) {
+        var operator = socialVulnIndex[m].operator;
+        var weight = socialVulnIndex[m].weight;
+        var name = socialVulnIndex[m].name;
+        allSVIThemes.push(name);
+        var tempChildren = socialVulnIndex[m].children;
+        var indicatorChildrenKey = [];
```
ptormene: indicatorChildrenKey was also defined outside. Please check the scopes. I also suggest to refactor the name into indicatorChildrenKeys

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
-    function showDuplicateMsg() {
-        $('#warning-duplicate').dialog('open');
+var indicatorChildrenKey = [];
+var allSVIThemes = [];
+var allPrimaryIndicators = [];
+var allRiskIndicators = [];
+
+function createIndexSimple(la, index) {
+    var indicator = [];
+    // setup the indicator with all the municipalities
```
ptormene: I suggest refactoring also the comments, using "zones" instead of "municipalities"

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
-
-    // Duplicate layer warnning message
-    function showDuplicateMsg() {
-        $('#warning-duplicate').dialog('open');
+var indicatorChildrenKey = [];
+var allSVIThemes = [];
+var allPrimaryIndicators = [];
+var allRiskIndicators = [];
+
+function createIndexSimple(la, index) {
```
ptormene: Please add a docstring specifying what are `la` and `index`

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
-    layerControl = L.control.layers(app.baseLayers);
-    map.panTo(new L.LatLng(39.399, -8.22));
-    map.setZoom(6);
-
-    // Duplicate layer warnning message
-    function showDuplicateMsg() {
-        $('#warning-duplicate').dialog('open');
+var indicatorChildrenKey = [];
+
+function createIndexSimple(la, index) {
```
ptormene: Please add a docstring to specify the meaning of `la` and `index`

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
+    var indicator = [];
+    // setup the indicator with all the municipalities
```
ptormene: Please change also the comments, where you still have `municipalities`

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
+    for (var ia = 0; ia < la.length; ia++) {
+        var temp = {};
+        temp.region = la[ia].properties[selectedRegion];
+        indicator.push(temp);
+        // districts is used inside of the d3 charts
+        // TODO change this to region
```
ptormene: Please check and refactor where you have `districts`

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
ptormene: From what I see afterwards, it looks like index is an array. I suggest to adopt naming conventions that give the reader a first-sight idea of the content (and cardinality) of your objects

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
-        });
-        $( '#econ-weight' ).val( $( '#slider-vertical' ).slider( 'value' ) );
-    });
-
-    // No Layer to remove warnning message
-    function showRemoveMsg() {
-        $('#warning-no-layer').dialog('open');
+    // Get the indicators children keys
+    for (var q = 0; q < index.length; q++) {
+        indicatorChildrenKey.push(index[q].name);
```
ptormene: Please check if you want the `name` or the `field` here

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
-                    var gridName = grid.split('/')[4];
-                    layerGrids.push(gridName);
+    // Match each primary indicator with it's respective data value
+    var primaryRiskIndicatorObj = {};
+    for (var ic = 0; ic < la.length; ic++) {
+        var tempObj = la[ic].properties;
+        for (var d = 0; d < indicatorChildrenKey.length; d++) {
+            for(var o in tempObj) {
+                if (indicatorChildrenKey[d] == o) {
+                    var tempValue = indicatorChildrenKey[d];
```
ptormene: Please use some more intuitive names for those temporary variables, so what you are doing here becomes simpler to read (and possibly add some comments)

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
+                    var tempValue2 = tempObj[o];
+                    var tempValue3 = tempObj[selectedRegion];
+                    if (primaryRiskIndicatorObj[tempValue] == undefined) {
+                        primaryRiskIndicatorObj[tempValue] = tempValue2;
+                        primaryRiskIndicatorObj.district = tempValue3;
```
ptormene: Please check all the `district` occurrences and refactor them

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
-                layersByCat[cat].push(layerTitle);
-                projectDefinition[name].push(pDef);
+    }
+    return primaryRiskIndicatorObj;
+}
+
+function createIndex(la, index) {
+    var ct = 0;
+    var indicator = [];
+    // setup the indicator with all the municipalities
```
ptormene: Please fix the comment

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
-                layerTitle = json[j].mapped_value;
-                layerNames[name].push(layerId);
-                layersByCat[cat].push(layerTitle);
-                projectDefinition[name].push(pDef);
+    }
+    return primaryRiskIndicatorObj;
+}
+
+function createIndex(la, index) {
+    var ct = 0;
```
ptormene: I don't understand this variable (`ct`). Is it used? Perhaps it could be removed?

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
+    var indicator = [];
+    // setup the indicator with all the municipalities
+    for (var ia = 0; ia < la.length; ia++) {
+        var temp = {};
+        temp.region = la[ia].properties[selectedRegion];
+        indicator.push(temp);
+    }
+    for (var i = 0; i < index.length; i++) {
+        for (var j = 0; j < la.length; j++, ct++) {
```
ptormene: You are incrementing `ct` but it looks like you are not using it

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
+            if (indicator[j].region == la[j].properties[selectedRegion]) {
+                var tempName = index[i].name;
```
ptormene: Please check if you need the `name` or the `field` here

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
-                layerId = json[j].id;
-                layerTitle = json[j].mapped_value;
-                layerNames[name].push(layerId);
-                layersByCat[cat].push(layerTitle);
-                projectDefinition[name].push(pDef);
+    }
+    return primaryRiskIndicatorObj;
+}
+
+function createIndex(la, index) {
```
ptormene: Please add a docstring also here, specifying the meaning of the parameters, and the return value. About the return value, it looks like you are building an indicator object and returning it, so you might refactor the function into createIndicator

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
-            selCat.appendChild(opt);
-            // Append layer list to dowpdown
-            var layerOpt = document.createElement('option');
+    }
+    return indicator;
+}
+
+function combineIndicators(nameLookUp, themeObj, JSONthemes) {
+    projectDef = sessionProjectDef;
+    var ct = 0;
```
ptormene: Also in this case, `ct` seems to be just incremented during the loops, but otherwise it's not used. If this is true, please remove all the unnecessary occurrences of this variable

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
-            opt.value = categoryTitle;
-            selCat.appendChild(opt);
-            // Append layer list to dowpdown
-            var layerOpt = document.createElement('option');
+    }
+    return indicator;
+}
+
+function combineIndicators(nameLookUp, themeObj, JSONthemes) {
+    projectDef = sessionProjectDef;
```
ptormene: What's the difference between the two global variables `projectDef` and `sessionProjectDef`? Are they both needed?

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
-        var layersArray = layersByCat[strUser];
-
-        for (var i in layersArray) {
-            var layers = layersArray[i];
-            var opt = document.createElement('option');
-            opt.innerHTML = layers;
-            opt.valuse = layers;
-            selLayer.appendChild(opt);
+    }
+    // first create an object with all of the district names
```
ptormene: I think all the occurrences of `district` and `region` should be refactored into the more generic `zone`

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
-            // Append layer list to dowpdown
-            var layerOpt = document.createElement('option');
+    }
+    return indicator;
+}
+
+function combineIndicators(nameLookUp, themeObj, JSONthemes) {
+    projectDef = sessionProjectDef;
+    var ct = 0;
+    var subIndex = {};
```
ptormene: is `subIndex` equivalent to `primaryIndicator`? I suggest using names such as: IRI, RI, SVI, riskIndicator, svIndicator, svTheme

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
-            opt.valuse = layers;
-            selLayer.appendChild(opt);
+    }
+    // first create an object with all of the district names
+    for (var t = 0; t < themeObj.length; t++) {
+        var tempRegion = themeObj[t].region;
+        subIndex[tempRegion] = 0;
+    }
+    // get some info aobut the themes
+    var themekeys = [];
```
ptormene: I suggest to use the camelCase convention also for `themeKeys`

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,864 @@
...
+        var themeName = JSONthemes[u].name;
+        var themeWeight = JSONthemes[u].weight;
+        themekeys.push(themeName);
+        themeWeightObj[themeName] = themeWeight;
+    }
+    // compute the subIndex values based on operator
+    if (operator == 'Simple sum (ignore weights)') {
+        for (var v = 0; v < themeObj.length; v++) {
+            var tempElementValue = 0;
+            var themeObjMunic = themeObj[v].region;
```
ptormene: Please refactor also these to zone

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+    //// Create the primary indicator objects ///
+    /////////////////////////////////////////////
+
+    var socialVulnIndex;
+    var riskIndex;
+
+    // Find all the nodes of type social vulnerability and risk index in the project def
+    for (var k in projectDef.children) {
+        if (projectDef.children[k].type == "Social Vulnerability Index") {
+            socialVulnIndex = projectDef.children[k].children;
```
ptormene: I would refactor `socialVulnIndex` into something like `svThemes`

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+        } else if (projectDef.children[k].type == "Risk Index") {
+            riskIndex = projectDef.children[k].children;
```
ptormene: I would refactor `riskIndex` into something like `riskIndicators`

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
-        })) {
-            tempAalWeight = pdData.weight;
-        }
-        (pdData.children || []).forEach(function(currentItem) {
-            findAalWeight(currentItem, [aal]);
-        });
-    }
+    // process each Social Vulnerability Index nodes
+    // Get all the primary indicators
+    var ct = 0;
```
ptormene: Again, check `ct`

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
 
-    function utfGridClickEvent(utfGrid) {
-        utfGrid.on('click', function(e) {
-            processIndicators(e);
-            selectedGrid = e;
-        });
+    for (var i = 0; i < socialVulnIndex.length; i++) {
```
ptormene: After the refactoring, it becomes clear that you are looping through the social vulnerability themes here.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+        for (var e = 0 ; e < socialVulnIndex[i].children.length; e++, ct++ ) {
```
ptormene: And here it becomes clear that you are looping through the svIndicators related to the current theme

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+            allPrimaryIndicators.push(socialVulnIndex[i].children[e].field);
```
ptormene: are primaryIndicators only including social vulnerability stuff, and not including risk stuff? If this is the case, please refactor into `allSvIndicators`

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
-            var pri = e.data.pri;
-
-            if (pri == undefined) {
-                pri = e.data.aal;
-            }
-            var munic_num = e.data['municipio'].split(',').length;
-            municipality = e.data['municipio'].split(',');
-
+    /////////////////////////////////////////
+    //// Compute the Category indicators ////
```
ptormene: Please refactor all occurrences of `Category` into `Theme`

ptormene: (and `cat` into `theme`)

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
-            /////////////////////////////////////////////
-            //// Create the primary indicator objects ///
-            /////////////////////////////////////////////
+    //build the category indicator object
+    var catData = [];
+    tempString = [];
 
-            for (var k in weightedPrimaryIndicator) {
-                delete weightedPrimaryIndicator[k];
+    function generateObject(temp) {
```
ptormene: Please change `temp` into something more intuitive, and possibly add a docstring describing it. Also `generateObject` sounds a bit generic. What kind of object are you generating?

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+        var munic = temp[0];
+        var theme = temp[1];
+        var value = parseFloat(temp[2]);
+        // add the theme and value to each category data object
+        for (var i = 0; i < catData.length; i++) {
+            if (catData[i].region == munic) {
```
ptormene: The structure of `catData` might be improved, to keep logically separated the theme values and the region. A possible solution would be to nest theme values one step further. So you would have `catData[i]` containing `region` (or `zone`) as the name of the zone, and `themeValues` as a dictionary theme -> value. In this way, you can iterate through the themeValues without checking every time if you are accessing the zone name.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+    }
 
-            for (var k in primaryIndicator) {
-                delete primaryIndicator[k];
-            }
+    // Find the theme information
+    for (var m = 0; m < socialVulnIndex.length; m++) {
+        var operator = socialVulnIndex[m].operator;
+        var weight = socialVulnIndex[m].weight;
+        var name = socialVulnIndex[m].name;
```
ptormene: Please be sure that you actually want the `name` and not the `field`

ptormene: (I think in this case `name` is correct, because themes aren't stored in actual fields in the layer)

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
-                for (var key in weightedPrimaryIndicator) {
-                    for (var i = 0; i < tempParentChildKey.length; i++) {
-                        tempCatSearchElements = parentChildKey[tempParentChildKey[i]]; //what we are looking for in weightedPrimaryIndicator
-                        var obj = weightedPrimaryIndicator[key];
-                        var piArray = []; // array of values from a single weightedPrimaryIndicator obj
-                        var mun = weightedPrimaryIndicator[key].municipality;
-
-                        for (var n = 0; n < tempCatSearchElements.length; n++) {
-                            piArray.push(obj[tempCatSearchElements[n]]);
+                var munic1 = la[o].properties[selectedRegion];
```
ptormene: Are those `1` needed?

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+                var theme1 = name;
+                tempString.push(munic1 + '|'+ theme1 +'|'+ tempSum);
+            } else if ( operator == "Weighted sum") {
+                for (var p2 in la[o].properties) {
```
ptormene: Are those `2` needed? Please check the scopes

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
-                        tempCIValues[tempParentChildKey[i]] = sum;
-                        tempCIValues["municipality"] = mun;
-
-                        // Multiply the category indicator data by the weighted value
-                        for (var k in tempCIValues) {
-                            for (var p in tempCatWeight) {
-                                if (k == p) {
-                                    tempCIValues[k] = (tempCIValues[p] * tempCatWeight[k]);
-                                }
+                var munic2 = la[o].properties[selectedRegion];
```
ptormene: Same as above

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+                var theme2 = name;
+                tempString.push(munic2 + '|'+ theme2 +'|'+ tempSum);
+            } else if ( operator == "Simple multiplication (ignore weights)") {
+                for (var p3 in la[o].properties) {
```
ptormene: Same as above

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
-                    }
-
-                    // Call the applyScaledValues method
-                    for (var l in scaledCatIndicator) {
-                        scaledCatIndicator[l][CIkeys[i]] = scaledCatIndicator[l].scaleCIvalues(l);
+                var munic3 = la[o].properties[selectedRegion];
+                var theme3 = name;
+                tempString.push(munic3 + '|'+ theme3 +'|'+ tempSum);
+            } else if ( operator == "Weighted multiplication") {
+                for (var p4 in la[o].properties) {
```
ptormene: Same as above

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
-
-            for (var k in weightedSviIndicator) {
-                delete weightedSviIndicator[k];
-            }
-
-            function newSVIvalues(SVIaverage, tempMunic) {
-                sviIndicator[tempMunic] = SVIaverage;
+                var munic4 = la[o].properties[selectedRegion];
+                var theme4 = name;
+                tempString.push(munic4 + '|'+ theme4 +'|'+ tempSum);
```
ptormene: Instead of building strings and then parse them, you might build an array of dictionaries. For instance, in each item of your array, you would have {'zone': 'Lisboa', 'theme': 'Population', 'value': 10.3}

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
-            var tempSviIndicator = {};
-            for (var i = 0; i < municipality.length; i++) {
-                tempSviIndicator[municipality[i]] = [];
-            }
-            tempSviSearchElements = Object.keys(weightedCatIndicator);
-            tempSviSearchElements.pop();
-            var sessionKey = Object.keys(weightedCatIndicator[tempSviSearchElements[0]]);
-            sessionKey.pop();
-            sessionKey.pop();
+    for (var p5 = 0; p5 < tempString.length; p5++) {
```
ptormene: Again, `5` is not necessary

ptormene: If you build a structured object as suggested, it would be much clearer here to get the information from that, instead of parsing the string.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+    var SVI = {};
+    var sviNameLookUp = 'SVI';
+    var sviJSONthemes = socialVulnIndex;
+    // SVI is an object with region and value
+    SVI = combineIndicators(sviNameLookUp, catData, sviJSONthemes );
 
-                if (index > -1) {
-                    sessionKey.splice(index, 1);
-                }
+    ///////////////
```
ptormene: Now that you have better factorized the scaling, you can remove these lines from here

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+    //// Compute the risk index ////
+    ////////////////////////////////
 
-                sviIndicator.SVIvalues(SVIsum, tempMunic);
-            };
+    createIndexSimple(la, riskIndex);
+    var riskIndicator = createIndex(la, riskIndex);
 
+    // capture all risk indicators for selection menu
+    for (var key in riskIndicator[0]) {
```
ptormene: A better structured riskIndicators object would let you avoid this `if`, because you would be browsing homogeneous data

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+    ////////////////////////////////////
+    //// Compute the Risk Indicator ////
+    ////////////////////////////////////
+
+    var RI = {};
+    var nameLookUp = 'RI';
+    var riJSONthemes = riskIndex;
+    RI = combineIndicators(nameLookUp, riskIndicator, riJSONthemes);
+
+    ///////////////
```
ptormene: Same as above

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+        }
+    } else if (iriOperator == "Weighted multiplication") {
+        for (var regionName in SVI) {
+            tempVal = (SVI[regionName] * sviWeight) * (RI[regionName] * riWeight);
+            IRI[regionName] = tempVal;
+        }
+    }
 
-            scaledSviIndicator = jQuery.extend(true, {}, sviIndicator);
+    ///////////////
```
ptormene: Same as above

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+        for (var key in RI) {
+            if (key == la[ix].properties[selectedRegion]) {
+                la[ix].newProperties['IR'] = (RI[key]).toFixed(5);
             }
+        }
 
-            for (var k in scaledAalIndicator) {
-                delete scaledAalIndicator[k];
+        for (var key in riskIndicator[ix]) {
+            if (riskIndicator[ix] != 'region') {
```
ptormene: Same as above, this `if` wouldn't be necessary if riskIndicator was better structured

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+                var tempThemeName = riskIndicator[ix][key];
+                la[ix].newProperties[key] = tempThemeName;
             }
+        }
 
-            for (var k in weightedAalIndicator) {
-                delete weightedAalIndicator[k];
+        for (var key in catData[ix]) {
+            if (key != 'region') {
```
ptormene: Same as above

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+function scale(IndicatorObj) {
+    var ValueArray = [];
+    var scaledValues = [];
+    for (var v in IndicatorObj) {
+        ValueArray.push(IndicatorObj[v]);
+    }
+    var tempMin = Math.min.apply(null, ValueArray),
+        tempMax = Math.max.apply(null, ValueArray);
+    for (var j = 0; j < ValueArray.length; j++) {
+        // TODO make sure not to devide by zero
```
ptormene: Please address this TODO

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+        geoJSONField: 'geometry',
+        layerOptions: {
+            fillOpacity: 1,
+            opacity: 1,
+            weight: 1,
+            stroke: true,
+            color: '#0000FF'
+        },
+        displayOptions: {},
+        title: "foobar"
```
ptormene: Please fix the title

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+            var x2js = new X2JS();
+
+            var jsonElement = x2js.xml_str2json(xmlText);
+            var featureType = jsonElement.WFS_Capabilities.FeatureTypeList.FeatureType;
+
+            // Find the SVIR keywords
+            var stringToLookFor = 'SVIR_QGIS_Plugin';
+            for (var i = 0; i < featureType.length; i++) {
+                if (featureType[i].Keywords.indexOf(stringToLookFor) > -1) {
+                    var tempTitle = featureType[i].Title;
```
ptormene: It looks like you are not using tempTitle and tempName, so they could be removed

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
-            //// Scale ////
-            ///////////////
+    // Get the layer metadata (project def)
+    $('#svir-project-list').change(function() {
+        $('#projectDef-spinner').show();
+        $('#iri-spinner').show();
+        // Get the selected layer
+        var selectedLayer = document.getElementById('svir-project-list').value;
+        // clean the selected layer to get just the layer name
+        selectedLayer = selectedLayer.substring(selectedLayer.indexOf("(") + 1);
```
ptormene: We have problems if the selected name contains a parenthesis. If you can fix it in this PR, it would be good. Otherwise please add a FIXME

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,850 @@
...
+                    '<p>This application was not able to get information about the selected layer</p>'
+                );
+            $('#ajaxErrorDialog').dialog('open');
             }
+        });
+    });
 
-            var tempKeys = Object.keys(iriIndicator);
 
-            for (var i = 0; i < t
```
ptormene: Please consider if these lines can be useful as a reference, or remove them

openquakeplatform/openquakeplatform/static/js/irv_viewer_PCP_category.js
```diff
@@ -33,13 +30,28 @@ function Category_PCP_Chart(catData, municipality, districName, concat) {
     var margin = {top: 100, right: 20, bottom: 10, left: 20},
         width = (winW - 200) - margin.left - margin.right,
         height = winH - margin.top - margin.bottom;
+    var tmpArray = [];
+    var array = [];
```
ptormene: Please rename tmpArray and array into something more intuitive

openquakeplatform/openquakeplatform/static/js/irv_viewer_PCP_category.js
```diff
@@ -33,13 +30,28 @@ function Category_PCP_Chart(catData, municipality, districName, concat) {
...
+    for (var i = 0; i < catData.length; i++) {
```
ptormene: Please add comments explaining what you are appending to array and tmpArray

openquakeplatform/openquakeplatform/static/js/irv_viewer_d3_tree.js
```diff
@@ -94,11 +94,43 @@
...
+                    // update the JSON with new weights
+                    traverse(sessionProjectDef, tempNewWeight);
+                }
+
                 nodeEnter.remove("text");
                 updateD3Tree(pdData);
             });
         }
 
+        function traverse(projectDef, tempNewWeight) {
```
ptormene: Please add a docstring explaining the purpose of this function

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: Please be aware that the risk and primary indicators have field elements, while the theme's do not, this is why you will find in some cases the use of 'field' and in others 'name'

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: The ct variable is important here, it prevents the for loop below from looping exponentially

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: subIndex is the child or SVI or IR. The issue with finding a good name is that in the case of SVI, it's children could be called 'themes'. While the children of IR I would consider to be something like 'risk indicators'. It is for this reason that I am using 'subIndex' as it is generic.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: I am not completely comfortable with the use of zone. For me region has more of a geographic context.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: done

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
ptormene: Please tell me in which line. I can't find where it's used.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: ct is used to prevent an exponential for loop

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: see above note

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: done

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
ptormene: In the past I had a discussion with Chris about what name we should use. We agreed that "zone" is a good choice because it is generic enough to avoid overlapping with other possible concepts. For instance, "region" could be confused with some sub-national zones, or with some multi-national named-regions. And "zone" is a word that is not unknown in the geographical context; for instance in QGIS there are some algorithms called something like "zonal statistics". For those reasons, after a thorough discussion with Chris, we decided to refactor all the code in the plugin to use "zone" instead of other possible candidates that we were using interchangeably. I believe it would be good to use a common naming convention in our tools. Nevertheless, if you feel more comfortable with the word "region", I am not strongly against it. Whatever you decide, just make sure you use one single convention in the whole web viewer, avoiding confusion with districts, municipalities, etc.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
ptormene: same as in the other comment

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: you can find it with a search for every assurance of '++ct'

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: yes, sessionProjectDef is the project definition as is was when uploaded from the QGIS tool, bit projectDef includes modified weights and is no longer the version that was uploaded from the QGIS tool.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
ptormene: I already saw that you are incrementing ct many times, but I can't find where you are actually using the value of ct (e.g., comparing it with something)

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
ptormene: Ok. Then, please, add a comment to clarify it

bwyss: done

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: Done

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,839 @@
...
+        var tempPropertiesObj = la[ic].properties;
+        // for each layer attribute, iterate over each indicator child key
+        for (var d = 0; d < indicatorChildrenKey.length; d++) {
+            // interate over each of the properties objects
+            for (var o in tempPropertiesObj) {
+                // check for a match between the indicator child key and the properties key
+                if (indicatorChildrenKey[d] == o) {
+                    // if a match is made, capture some information about the feature
+                    // and populate the primary risk indicator object
+                    var tempValue = indicatorChildrenKey[d];
```
ptormene: The new comments look good. Now we need to rename tempValueN into something that describes their meaning.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,839 @@
...
+                    var tempValue2 = tempPropertiesObj[o];
+                    var tempValue3 = tempPropertiesObj[selectedRegion];
+                    if (primaryRiskIndicatorObj[tempValue] == undefined) {
+                        primaryRiskIndicatorObj[tempValue] = tempValue2;
+                        primaryRiskIndicatorObj.district = tempValue3;
```
ptormene: Please check all occurrences of district and refactor them into region (if you chose region rather than zone)

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,839 @@
...
-    layerControl = L.control.layers(app.baseLayers);
-    map.panTo(new L.LatLng(39.399, -8.22));
-    map.setZoom(6);
-
-    // Duplicate layer warnning message
-    function showDuplicateMsg() {
-        $('#warning-duplicate').dialog('open');
+var indicatorChildrenKey = [];
+
+function createIndexSimple(la, riskIndicators) {
```
ptormene: I've just noticed you have both `createIndexSimple` and `createIndex`. Can you write somewhere what's the difference and in which cases they are used?

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,839 @@
...
-
-                if (hasGrid == true) {
-                    gridList = 1;
-                    utfGrid = new L.UtfGrid(TILESTREAM_URL+
-                        selectedLayer+
-                        '/{z}/{x}/{y}.grid.json?callback={cb}', {Default: false, JsonP: false});
-                    map.addLayer(utfGrid);
-                    utfGridClickEvent(utfGrid);
+    } else if (operator == 'Weighted sum') {
+        for (var v1 = 0; v1 < themeObj.length; v1++) {
```
ptormene: I guess a previous comment was hidden by github after some changes in the code around it. I don't see the point of using a different name for the variables in completely separated loops (v1, v2, vN). Couldn't you just use v instead?

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,839 @@
...
-            if(pdTempPILevel.hasOwnProperty(prop)) {
-                if(pdTempPILevel[prop] === foo) {
-                    return prop;
-                }
+    //build the theme indicator object
+    var themeData = [];
+    indicatorString = [];
+
+    function generateObject(indicatorArray) {
+        var munic = indicatorArray[0];
```
ptormene: Please check all occurrences of munic and replace them with region

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,839 @@
...
-        for(var prop in pdTempPILevel) {
-            if(pdTempPILevel.hasOwnProperty(prop)) {
-                if(pdTempPILevel[prop] === foo) {
-                    return prop;
-                }
+    //build the theme indicator object
+    var themeData = [];
+    indicatorString = [];
+
+    function generateObject(indicatorArray) {
```
ptormene: `Object` sounds a bit generic. Please rename it into something more intuitive.
Furthermore, it looks a bit strange that you have no return value, and instead you manipulate a global `themeData` object. Please add a comment to clarify these aspects.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,839 @@
...
-                                var pos;
-                                pos = parentChildKey[ep].indexOf(elementName);
-                                if (!~pos) {
-                                    parentChildKey[ep].push(elementName);
-                                }
-                            }
+        // iterate over the layerAttributes to access the data
+        for (var o = 0; o < la.length; o++) {
+            var tempSum = 0;
+            var munic = la[o].properties[selectedRegion];
```
ptormene: munic -> region

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,839 @@
...
-            // Multiply the copy of primary indicator data by the weighted value
-            for(var p1 in weightedPrimaryIndicator) {
-                for(var p2 in weightedPrimaryIndicator[p1]) {
-                    if(tempWeight.hasOwnProperty(p2)) {
-                        weightedPrimaryIndicator[p1][p2] *= tempWeight[p2];
+                // Grab the average
+                var average = tempSum / tempIndicatorChildrenKeys.length;
+                indicatorString.push(munic + '|'+ theme +'|'+ average);
+            } else if ( operator == "Simple sum (ignore weights)") {
+                for (var p1 in la[o].properties) {
```
ptormene: Same as above, about unnecessary p1, p2, ... pN

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,839 @@
...
-                sviIndicator[tempMunic] = SVIaverage;
-            }
-
-            // Associate the newSVIvalues method with the svi object
-            sviIndicator.SVIvalues = newSVIvalues;
-            var tempSviParentChildKey = [];
-            var tempSviIndicator = {};
-            for (var i = 0; i < municipality.length; i++) {
-                tempSviIndicator[municipality[i]] = [];
+                indicatorString.push(munic + '|'+ theme +'|'+ tempSum);
```
ptormene: If you had an array such as indicatorInfo = [],
then you can do something like
indicatorData.push({"region": region, "theme": theme, "sum": sum})
And afterwards you can access it easily without parsing a string.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,839 @@
...
+    var scaledValues = [];
+    for (var v in IndicatorObj) {
+        ValueArray.push(IndicatorObj[v]);
+    }
+    var tempMin = Math.min.apply(null, ValueArray),
+        tempMax = Math.max.apply(null, ValueArray);
+    for (var j = 0; j < ValueArray.length; j++) {
+        // make sure not to devide by zero
+        if (tempMax  == tempMin) {
+            ValueArray[j] = 1;
```
ptormene: Please add a comment to specify that `1` is an arbitrary choice to translate a flat array into an array where each element equals to 1 (we might ask Chris if he likes such decision or if he prefers 0, or 0.5, or else instead).

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,839 @@
...
-            ///////////////
+    // Get the layer metadata (project def)
+    $('#svir-project-list').change(function() {
+        $('#projectDef-spinner').show();
+        $('#iri-spinner').show();
+        $('#regionSelectionDialog').empty();
+        // Get the selected layer
+        var selectedLayer = document.getElementById('svir-project-list').value;
+        // clean the selected layer to get just the layer name
+        selectedLayer = selectedLayer.substring(selectedLayer.indexOf("(") + 1);
```
ptormene: Please add a FIXME comment here, specifying that this doesn't work in case the project title contains any parentheses

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,799 @@
...
+var indicatorChildrenKey = [];
+
+
+function createIndex(la, index) {
+    var indicator = [];
+    // setup the indicator with all the regions
+    for (var ia = 0; ia < la.length; ia++) {
+        var temp = {};
+        temp.region = la[ia].properties[selectedRegion];
+        indicator.push(temp);
```
ptormene: `indicator.push({'region': la[ia].properties[selectedRegion]});`

ptormene: Or, rather:
var region = la[ia].properties[selectedRegion];
indicator.push({'region': region});
regions.push(region);

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,799 @@
...
-            sessionPrimaryIndicator = JSON.parse( JSON.stringify( primaryIndicator ) );
-
-            // Multiply the copy of primary indicator data by the weighted value
-            for(var p1 in weightedPrimaryIndicator) {
-                for(var p2 in weightedPrimaryIndicator[p1]) {
-                    if(tempWeight.hasOwnProperty(p2)) {
-                        weightedPrimaryIndicator[p1][p2] *= tempWeight[p2];
+                // Grab the average
+                var average = tempSum / tempIndicatorChildrenKeys.length;
+                indicatorString.push(region + '|'+ theme +'|'+ average);
```
ptormene: (This comment was automatically hidden by github after you pushed some changes on near code)
If you had an array such as indicatorInfo = [],
then you can do something like
indicatorData.push({"region": region, "theme": theme, "sum": sum})
And afterwards you can access it easily without parsing a string.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,799 @@
...
-                for (var key in weightedPrimaryIndicator) {
-                    for (var i = 0; i < tempParentChildKey.length; i++) {
-                        tempCatSearchElements = parentChildKey[tempParentChildKey[i]]; //what we are looking for in weightedPrimaryIndicator
-                        var obj = weightedPrimaryIndicator[key];
-                        var piArray = []; // array of values from a single weightedPrimaryIndicator obj
-                        var mun = weightedPrimaryIndicator[key].municipality;
-
-                        for (var n = 0; n < tempCatSearchElements.length; n++) {
-                            piArray.push(obj[tempCatSearchElements[n]]);
+                indicatorString.push(region + '|'+ theme +'|'+ tempSum);
```
ptormene: As above

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,799 @@
...
-                        tempCIValues[tempParentChildKey[i]] = sum;
-                        tempCIValues["municipality"] = mun;
-
-                        // Multiply the category indicator data by the weighted value
-                        for (var k in tempCIValues) {
-                            for (var p in tempCatWeight) {
-                                if (k == p) {
-                                    tempCIValues[k] = (tempCIValues[p] * tempCatWeight[k]);
-                                }
+                indicatorString.push(region + '|'+ theme +'|'+ tempSum);
```
ptormene: As above

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,799 @@
...
-
-                    // Scale the values
-                    for (var j = 0; j < tempCIvalues.length; j++) {
-                        scaleCIvalues.push( (tempCIvalues[j] - tempCImin) / (tempCImax - tempCImin) );
-                    }
-
-                    // Call the applyScaledValues method
-                    for (var l in scaledCatIndicator) {
-                        scaledCatIndicator[l][CIkeys[i]] = scaledCatIndicator[l].scaleCIvalues(l);
+                indicatorString.push(region + '|'+ theme +'|'+ tempSum);
```
ptormene: As above

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,799 @@
...
-                tempCIvalues = [];
-            }
-
-            // The data passed into d3 need to be an array of objects
-            // Place the category indicator objects into an array
-            var catData = [];
-
-            for (var k in scaledCatIndicator) {
-                catData.push(scaledCatIndicator[k]);
+                indicatorString.push(region + '|'+ theme +'|'+ tempSum);
```
ptormene: As above

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,799 @@
...
-
-                var SVIsum = 0
-                $.each(tempArray,function() {
-                    SVIsum += this;
-                });
+    for (var p5 = 0; p5 < indicatorString.length; p5++) {
+        // capture an array for each record
+        var indicatorArray;
+        indicatorArray = indicatorString[p5];
+        indicatorArray = indicatorArray.split("|");
```
ptormene: Here, instead of parsing a string, you can get the dictionary from the list and have all the elements easily available.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,799 @@
...
-
-            if (cat != undefined && type == 'svir-qgis') {
-                layerId = json[j].id;
-                layerTitle = json[j].mapped_value;
-                layerNames[name].push(layerId);
-                layersByCat[cat].push(layerTitle);
-                projectDefinition[name].push(pDef);
-            }
+    }
+    // first create an object with all of the district names
```
ptormene: Please check all occurrences of district also in the comments

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -14,1043 +14,799 @@
...
-        for(var prop in pdTempPILevel) {
-            if(pdTempPILevel.hasOwnProperty(prop)) {
-                if(pdTempPILevel[prop] === foo) {
-                    return prop;
-                }
+    //build the theme indicator object
+    var themeData = [];
+    indicatorString = [];
+
+    function generateThemeObject(indicatorArray) {
```
ptormene: Instead of accepting an array, you can pass the dictionary, or region, theme and sum separately.

ptormene:
LGTM!

PR #380, opened 2015-03-09
Move git_suffix out of utils.py since utils requires Django.

daniviga:
When the code is installed in a clean environment, without Django,
utils.py cannot be imported but is required to get code version.


nastasi:
LGTM

PR #382, opened 2015-03-16
Fix migrations copy in deploy.sh

daniviga:
Fix migrations copy in deploy.sh where some file extensions are not present in the src folder.

`cp` fails if any of `*.sh`, `*.sql` or `*.py` is not present in the migrations dir. This will bring `deploy.sh` to a failure.


nastasi-oq:
Never happens because one migration appear in the first version of migration enabled version.
In any case this fix is more consistent so LGTM.

daniviga:
This happened to me just few minutes ago.

daniviga:
Test:
```bash
$ mkdir s d
$ touch s/a.py s/b.sh
$ cp s/*.{py,sh,sql} d/
cp: cannot stat ‘s/*.sql’: No such file or directory
```

PR #383, opened 2015-03-16
fix per-user development stop target

nastasi-oq:
list process of current-user only to check if all target process instances are shutted down.


daniviga:
LGTM

PR #386, opened 2015-03-17
Irv refactor

bwyss:
This PR resolves a small issue found in the IRV application. When the user makes a indicator selection for the map, then changes a weight, the selection is set back to 'IRI'.


daniviga:
LGTM

PR #387, opened 2015-03-18
Irv negative weights

bwyss:
This PR updates the style of the tree chart to match the QGIS tool (positive and negative weights)
These changes can been seen/test on the dev server

### DIFF
--- a/openquakeplatform/openquakeplatform/static/js/irv_viewer_d3_tree.js
+++ b/openquakeplatform/openquakeplatform/static/js/irv_viewer_d3_tree.js
@@ -68,8 +68,8 @@
 
             $(function() {
                 $("#spinner-"+id).width(100).spinner({
-                    min: 0,
-                    max: -100,
+                    min: -100,
+                    max: 100,
                     step: 0.000001,
                     numberFormat: "n",
                 });
--- a/openquakeplatform/openquakeplatform/static/js/irv_viewer_d3_tree.js
+++ b/openquakeplatform/openquakeplatform/static/js/irv_viewer_d3_tree.js
@@ -274,12 +274,7 @@
                 .style("fill", function(d) {
                     if (d.operator != undefined) {
                         // Check for operators that ignore weights and style accordingly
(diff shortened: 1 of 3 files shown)


openquakeplatform/openquakeplatform/static/js/irv_viewer_d3_tree.js
```diff
@@ -270,7 +284,61 @@
...
+                            ignoreWeightsStr = '(' + parts[1];
+                        }
+                        return ignoreWeightsStr;
+                    }
+                })
+                .style("fill", function(d) {
+                    if (d.operator != undefined) {
+                        // Check for operators that ignore weights and style accordingly
+                        var color;
+                        if (d.operator.indexOf('ignore') != -1) {
```
ptormene: On my side I removed this check, because ignoreWeightStr is empty if the operator does not ignore weights. So, here I use color = '#660000' always.

ptormene:
We have 2 differences in our tools:
1) on my side, the name of the operator is black, and only the (ignore weights) part is in a different color;
2) when you click on the operator, you should be able to set the weights of the nodes that are combined by that operator. Instead, in the web viewer you set the weight of the same node where the operator is defined (and not of its children)

ptormene:
You might also add the red color to the weight percentage, in case the weight is ignored (I've added that in https://github.com/gem/oq-svir-qgis/commit/3c667d487f6eb958d6988ed379d147bdf965b2fc)

bwyss:
I have addressed point 1 and 2. I agree with point two. While there are server labels on the d3.js tree, the only element in the web application that can be modified are the weights. It is for this reason the only clickable element in the chart should be the weight labels IMHO.

ptormene:
I agree. If Chris doesn't require to change also the operators, then you can remove the onclick from the operator label.

ptormene:
Let's wait a little further before merging. We need to discuss some things with Chris, decide some specifications and align the plugin and your tool.

PR #388, opened 2015-03-18
Layers selector

bwyss:
This PR provides logic to hide the leaflet layer controller when no layers are loaded, and then show the controller when layers are present. The feature is only required in the Hazus and HRDE applications as all other leaflet based applications are preloaded with layers.


daniviga:
LGTM. Maybe in the future we can think about moving the shared code (i.e. ` function checkLayerController() {`) in a common js library.

PR #390, opened 2015-03-20
Improve traceback display. Use bootstrap modal instead of popups

daniviga:
This solves the issue where the same traceback is reported to all the failing jobs, causing confusion on the causes of a failure.
This also replace pop-ups with a modal dialog. Pop-ups are evil: sometime they are blocked, they go back the main windows... and so; since our pop-ups are "self called" (they are not pointing to an external url) are also difficult to style and manage. Now syntax highlighting can be easily added.

screenshot from 2015-03-20 18 14 25

Fixes: https://bugs.launchpad.net/oq-platform/+bug/1434177

More testing is needed before releasing the [WIP].

### DIFF
--- a/openquakeplatform/openquakeplatform/icebox/static/icebox/calculate.js
+++ b/openquakeplatform/openquakeplatform/icebox/static/icebox/calculate.js
@@ -148,6 +148,11 @@
      });
    var outputs = new Outputs();
 
+   var refresh_calcs;
+
+   function setTimer() {
+       refresh_calcs = setInterval(function() { calculations.fetch({reset: true}) }, 5000);
+   }
 
    /* classic event management */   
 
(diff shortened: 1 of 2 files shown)


openquakeplatform/openquakeplatform/icebox/static/icebox/calculate.js
```diff
@@ -191,5 +196,14 @@
                               alert(xhr.responseText);
 		                        }});
                       });
+      $(document).on('click', 'button[id=open_trace]',
+                function(e) {
+                        clearInterval(refresh_calcs);
+                      });
+      $(document).on('hidden.bs.modal', 'div[id^=traceback]',
```
nastasi-oq: ```traceback-``` instead of ```traceback``` is more consistent.

nastasi-oq:
LGTM

PR #391, opened 2015-03-25
Simplify django site configuration

daniviga:
GeoNode already provides an helper to do that. This PR removes redundant code.

See: https://github.com/gem/geonode/blob/2.0.x/geonode/base/management/commands/fixsitename.py


nastasi-oq:
LGTM

PR #397, opened 2015-04-14
taxtweb: enable input configuration using part of url

nastasi-oq:
A confortable way to test it is looping between application and the permalink at the bottom of the page.
1. open taxtweb and configure a taxonomy
2. open permalink in a new window
3. compare the original taxtweb page with the new
4. change some parameter
5. back to step n°2


daniviga:
In `index.html`:

```html
<meta http-equiv="Content-Type" content="text/html; charset=iso-8859-15">
```

Please use UTF-8 instead of ISO

daniviga:
In `index.html`: please fix indentation of some css lines (i.e. https://github.com/gem/oq-platform/blob/taxt-bidi/openquakeplatform/openquakeplatform/taxtweb/templates/taxtweb/index.html#L60)

daniviga:
In `index.html`, line 294: there's a double quote

daniviga:
In `index.html`, line 325: orphan `<tr>`

openquakeplatform/openquakeplatform/taxtweb/views.py
```diff
@@ -29,43 +29,107 @@
...
+    }
+    //
+    //  MATERIAL (Dir X)
+
+
+
+}
+    """
+
+    return True
```
daniviga: Probably this can be remove entirely

openquakeplatform/openquakeplatform/taxtweb/views.py
```diff
@@ -29,43 +29,107 @@
...
 
+    if 'HTTP_HOST' in request.META:
+        proto = (request.META['HTTP_X_FORWARDED_PROTO'] if
+                 'HTTP_X_FORWARDED_PROTO' in request.META else 'http')
+        if request.META['HTTP_HOST'].startswith('taxtweb'):
+            taxt_prefix = proto + '://' + request.META['HTTP_HOST']
+        else:
+            taxt_prefix = proto + '://' + request.META['HTTP_HOST'] + '/taxtweb'
+    else:
+        taxt_prefix = "http://taxtweb.openquake.org"
```
daniviga: For the future I would suggest to move this value in the settings. I would prefer to have another configuration parameter rather then an hardcoded url in the code.

daniviga:
LGTM

PR #400, opened 2015-04-15
SVIR: new API to append a project definition to an existing layer

ptormene:
Given a layer identifier and a project definition, if the logged user is enabled to modify the layer, the project definition will be added to the layer's supplemental information. In order to ensure backward compatibility, the supplemental information can contain a single json object. In that case, it will be converted into an array, and both the existing project definition and the new one will be sequentially appended.


daniviga:
LGTM if @bwyss tests pass

PR #402, opened 2015-04-22
SVIR - Fix the API to add a new project definition to one of the available layers

ptormene:
This looks too lucky to be true, but Ben confirmed it works. I found out that `layer` has the attribute `supplemental_information`, so we don't need to retrieve that through the `base_resourcebase` table. I did not expect it to be accessible directly from `layer`, so in the previous implementation I looked for that inside the metadata stuff.
After saving `layer`, everything is automatically synchronized, so the new project definition is promptly available, without needing to trigger any further updating procedure.


bwyss:
LGTM

daniviga:
LGTM

PR #403, opened 2015-04-23
SVIR - Stream socioeconomic data

ptormene:
Before this PR, the exported csv was first built, then returned. Now I use a generator to stream the data. This will be helpful to improve the QGIS plugin's downloader, which will be able to start streaming the data without the initial waiting.
No companion PR at QGIS side is required (these changes do not break the current downloader)


micheles:
LGTM

PR #406, opened 2015-05-05
Update profile

bwyss:
When a user access the permission dialog, they are reminded to update their profile if it does not include a user name, user organization or user position.

### DIFF
--- a/openquakeplatform/openquakeplatform/templates/_permissions.html
+++ b/openquakeplatform/openquakeplatform/templates/_permissions.html
@@ -3,15 +3,15 @@
   <div class="control-group" id="permission-view-download" >
 
 
-    {% if request.user.get_profile.organization_name = 'undefined' or request.user.get_profile.country = None %}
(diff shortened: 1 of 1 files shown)


openquakeplatform/openquakeplatform/templates/_permissions.html
```diff
@@ -1,7 +1,19 @@
 {% load i18n %}
 <div class="modal-body" id="permissions-body" >
   <div class="control-group" id="permission-view-download" >
-    <label> By making this item available to other OpenQuake Platform users you acknowledge that you have read and agree the terms of use.</label>
+
+
+    {% if request.user.get_profile.name = None or request.user.get_profile.name = '' or request.user.get_profile.organization = None or request.user.get_profile.position = None or request.user.get_profile.position = '' or request.user.get_profile.position = '' %}
```
daniviga: ```request.user.get_profile.position``` is repeated twice at the end.

daniviga:
LGTM

PR #409, opened 2015-05-11
Leaflet control orderlayers

bwyss:
This PR provides the user with that ability to change the order of the layers.
The Layer order leaflet plugin does not manage the leaflet mouse and click events that are associated with the UTFGrids. Because of the we need to manage the leaflet click and mouse event index objects. This is done each time a layer is added to the application or when the user changes the layer order.

This can be tested by adding multiple layers into the HRDE application and checking that the UtfGrid click event corresponds to the layer that is at the top of the layer controller.


daniviga:
Is it possible to hide the "Layer manager" box when no layers are loaded? It's confusing, because it's empty but when you hover no more than "layer manager" title appears

daniviga:
I don't think it's releated to this PR, but I get

```Uncaught ReferenceError: capitalize is not defined``` when I click a UTF grid (i.e. the hazard_CAN10-WESTH_input_asc_area layer)

**Comment by daniviga** [comment]
Another thing that can be misleading:
1) add layer "A"
2) add layer "B"

Move layer B on top, deactivate it and keep only layer "A" click on the map: you'll get data from layer "B" which is at the top, but is off and hidden

File: openquakeplatform/openquakeplatform/static/leaflet-control-orderlayers/dist/css/leaflet.control.orderlayers.css [hunk 1]
```diff
@@ -0,0 +1,51 @@
...
+
+.leaflet-control-layers .leaflet-row .leaflet-down {
+	float: left;
+	background-image: url(../images/arrow-down.png);
+	background-repeat: no-repeat;
+	background-position: center;
+    height: 20px;
+    width: 16px;
+    opacity: 0.75;
+    cursor: pointer;
```
daniviga: Please fix the indentation (also for lines above)

openquakeplatform/openquakeplatform/static/css/hrde.css
```diff
@@ -132,6 +132,26 @@ a {
   width: 100%;
 }
 
+.leaflet-control-layers {
+  width: 350px;
+}
+
+.leaflet-control {
+  clear: none;
+}
```
daniviga: Adding `overflow: hidden` you can avoid scroll bars in the control div:

From
screenshot from 2015-05-15 15 33 19
To
screenshot from 2015-05-15 15 32 58

bwyss:
I will need more time to find a solution to the issues when the top layer in the layer controller is deactivated.

daniviga:
Ok, let's open a bug, we'll address it later. LGTM

bwyss:
here is the bug: https://bugs.launchpad.net/oq-platform/+bug/1455510

PR #412, opened 2015-05-14
Taxt short

nastasi-oq:
allow input of taxonomy as sub-path of .../taxonomy URL
full or short taxonomy version are accepted
a new taxonomy type selector is present at bottom of the app
the permalink in vulnerability now load the taxonomy into the taxt popup window


openquakeplatform/openquakeplatform/taxtweb/test/config.py
```diff
@@ -0,0 +1 @@
+basepath="http://oq-platform2-mn.gem.lan:8000/taxtweb/"
```
daniviga: Please replace with `http://localhost:8000`

daniviga:
Apart from very minor comment, LGTM

daniviga:
Great, LGTM

PR #413, opened 2015-05-15
Irv median line

bwyss:
Include a median line in the IRI and Primary charts.

Once a project is loaded, you can open and view each of the three PCP charts. 
The charts now include a median line that is gray and thicker then all the others.

It should look something like this:
screen shot 2015-05-15 at 2 42 15 pm

(user documentation: http://www.globalquakemodel.org/openquake/support/documentation/platform/irv/ )


ptormene:
When I click "Load Project", the dialog can be slightly improved:
* I would change the title from "Load Project Definition" into "Load Project"
* I would remove the label "Load Integrated Risk Project" and keep just "Filter the project list"

ptormene:
Apart from the comment, which probably is not related to this PR (but I suggest to fix it here, since it's a small thing), LGTM

PR #416, opened 2015-05-25
Remove the bluemarble baselayer from previously created maps

daniviga:
Bluemarble has been already deprecated as baselayer for new maps, but it's still present in fixtures and pre-generated maps (and their child).

This PR updates pre-generated maps fixures and add an SQL migration to remove the layers from any maps already stored.


nastasi-oq:
LGTM

PR #417, opened 2015-05-25
Forbid adding a new project definition if it has the same title of an old one

ptormene:
Also avoid adding the project definition if the title is undefined or blank


bwyss:
LGTM

PR #419, opened 2015-05-26
Add the unaccent custom type (both case insensitive and accent insensitive)

ptormene:
Addresses https://bugs.launchpad.net/oq-platform/+bug/1459231

### DIFF
--- a/openquakeplatform/migrations/003_add_icmp_unaccent_operator.sql
+++ b/openquakeplatform/migrations/003_add_icmp_unaccent_operator.sql
@@ -1,0 +1,19 @@
+DROP EXTENSION IF EXISTS unaccent;
+DROP OPERATOR IF EXISTS =~@ (character varying, character varying);
+DROP FUNCTION IF EXISTS icompare_unaccent(character varying, character varying);
+
+CREATE EXTENSION unaccent;
+CREATE FUNCTION icompare_unaccent(character varying, character varying) RETURNS boolean
+    AS 'SELECT lower(unaccent($1)) LIKE lower(unaccent($2));'
+    LANGUAGE SQL
+    IMMUTABLE
+    RETURNS NULL ON NULL INPUT;
+
+CREATE OPERATOR =~@ (
+    LEFTARG = character varying,
+    RIGHTARG = character varying,
+    PROCEDURE = icompare_unaccent,
+    COMMUTATOR = =~@,
+    NEGATOR = !=~@
+);
+
--- a/openquakeplatform/migrations/003_add_icmp_unaccent_operator.sql
+++ b/openquakeplatform/migrations/003_add_icmp_unaccent_operator.sql
@@ -1,6 +1,6 @@
-DROP EXTENSION IF EXISTS unaccent;
 DROP OPERATOR IF EXISTS =~@ (character varying, character varying);
 DROP FUNCTION IF EXISTS icompare_unaccent(character varying, character varying);
+DROP EXTENSION IF EXISTS unaccent;
 
 CREATE EXTENSION unaccent;
 CREATE FUNCTION icompare_unaccent(character varying, character varying) RETURNS boolean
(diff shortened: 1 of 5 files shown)


openquakeplatform/openquakeplatform/models.py
```diff
@@ -0,0 +1,68 @@
...
+    Custom CharField that can be filtered using the 'unaccent' custom
+    lookup criterion. It leverages the custom '=~@' postgres operator,
+    that uses as procedure the custom 'icompare_unaccent' function.
+    This fieldtype can therefore be filtered both case insensitive
+    and accent insensitive.
+    """
+    def get_db_prep_lookup(self, lookup_type, value,
+                           connection, prepared=False):
+        if lookup_type == 'unaccent':
+            return ["%%%s%%" % connection.ops.prep_for_like_query(value)]
```
micheles: Can you add a comment here? Which SQL query is generated? I guess
`filter(authors__unaccent=author)` becomes something like `authors =~@ author`

openquakeplatform/openquakeplatform/vulnerability/views.py
```diff
@@ -117,7 +117,7 @@ def list_entries(request, **kwargs):
     if llrs:
         geninfo = geninfo.filter(llrs=llrs)
     if author:
-        geninfo = geninfo.filter(authors__icontains=author)
+        geninfo = geninfo.filter(authors__unaccent=author)
```
micheles: This is the ONLY place where the unaccent feature is used? If yes, instead of creating a monster perhaps it would have been possible to make a special case for this, use a raw query with something like

`select ... from table where unaccent(author)=unaccent(%s)`

openquakeplatform/openquakeplatform/models.py
```diff
@@ -0,0 +1,68 @@
...
+                       'startswith', 'istartswith', 'endswith', 'iendswith'):
+        lookup = "%s::text"
+
+    # Use UPPER(x) for case-insensitive lookups; it's faster.
+    if lookup_type in ('iexact', 'icontains', 'unaccent',
+                       'istartswith', 'iendswith'):
+        lookup = 'UPPER(%s)' % lookup
+
+    return lookup
+
```
micheles: If in the next release of Django lookup_cast changes in an incompatible way, you may have mysterious errors :-(

openquakeplatform/openquakeplatform/vulnerability/views.py
ptormene: I believe this feature might be useful in general, so it was worth investigating and creating a custom field that can be reused whenever we want. For instance, I wouldn't be surprised if Cata, after discovering this issue for that special field, would ask us to implement the same for other fields.

openquakeplatform/openquakeplatform/models.py
ptormene: But fortunately in next releases Django natively provides the accent-insensitive filtering, so we will remove our customization and use the standard way.

openquakeplatform/migrations/003_add_icmp_unaccent_operator.sql
```diff
@@ -0,0 +1,19 @@
+DROP OPERATOR IF EXISTS =~@ (character varying, character varying);
+DROP FUNCTION IF EXISTS icompare_unaccent(character varying, character varying);
+DROP EXTENSION IF EXISTS unaccent;
+
+CREATE EXTENSION unaccent;
+CREATE FUNCTION icompare_unaccent(character varying, character varying) RETURNS boolean
+    AS 'SELECT lower(unaccent($1)) LIKE lower(unaccent($2));'
+    LANGUAGE SQL
```
micheles: Notice that at the Django level you are doing an UPPER

openquakeplatform/migrations/003_add_icmp_unaccent_operator.sql
```diff
@@ -0,0 +1,19 @@
...
+    IMMUTABLE
+    RETURNS NULL ON NULL INPUT;
+
+CREATE OPERATOR =~@ (
+    LEFTARG = character varying,
+    RIGHTARG = character varying,
+    PROCEDURE = icompare_unaccent,
+    COMMUTATOR = =~@,
```
micheles: Please check if both `A =~@ B` and` B =~@ A` are the same or not.

micheles:
LGTM, but please add a comment about the commutativity of the operator

PR #420, opened 2015-05-29
Irv small ux improvements

bwyss:
This PR includes several small UX improvements. 

1. Automatically select and trigger the selection of the first project definition in the selection dropdown menu. To test, open the IRV application, and select a project. The loading spinner will indicate that the APIs are working, after a few moments, the loading spinner will be hidden and now you can see that the first project definition has been automatically selected in the dropdown menu and triggered. In addition, this d3.js project definition tree is now rendered.

2. Automatically select and trigger the selection of the first primary indicator. Similar to #1, once a project has been selected, the first primary indicator from inside the 'primary indicator chart' tab is automatically selected and triggered.

3. Disable the theme tab if there is only one theme in themeData. This can be tested by loading a project that has only one SVI theme child. You will now see that once the project is loaded, the 'SVI theme chart' tab will be disabled.

4. If the selected zone includes numbers instead of chericters, then these values were being used to identify the max value used for the SVI theme chart. To correct this I am now discriminating againset the 'region' elements.

5. Provide a default operator for the pd tree (for corner cases)

6. Call the project list API each time the load project dialog is opened. This will allow the user to access newly created project without reloading the page.

### DIFF
--- a/openquakeplatform/openquakeplatform/static/js/irv_viewer.js
+++ b/openquakeplatform/openquakeplatform/static/js/irv_viewer.js
@@ -907,6 +907,14 @@ var startApp = function() {
                     );
                 }
                 $('#projectDef-spinner').hide();
+
+                // select the first indicator
+                var menuOption = $('#pdSelection');
+                menuOption[0].selectedIndex = 1;
+                // trigger first indicator
+                setTimeout(function() {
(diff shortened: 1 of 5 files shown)


ptormene:
Other issues can be addressed in separate PRs. For what concerns the topic of this PR, LGTM

PR #428, opened 2015-06-09
Taxt short2

nastasi-oq:
All numerical fields (are all in "Building Information" tab) now are verified in 2 different way:
 - interactively if you compile the form
 - not interactively if you are opening a taxtonomy passing it by part of a URL

**HOW TO TEST IT**
 - open taxtweb and play with "Building Information" tab
 - click on "Permalink" when no field is broken and try to manipulate the resulting new URL
 - placing strings instead of numbers, negative number, float instead of integer and so on.

### DIFF
--- a/openquakeplatform/openquakeplatform/taxtweb/static/taxtweb/js/taxtweb.js
+++ b/openquakeplatform/openquakeplatform/taxtweb/static/taxtweb/js/taxtweb.js
@@ -2591,6 +2591,10 @@ function taxt_BuildTaxonomy()
             $('#noStoreysE12').addClass('gem_field_alert');
             h12 = false;
         }
+        else if (parseInt($('#noStoreysE11').val()) == parseInt($('#noStoreysE12').val())) {
+            ret_s.s = "Number of storey above ground: invalid range.";
+            return (false);
+        }
         else {
             $('#noStoreysE12').removeClass('gem_field_alert');
         }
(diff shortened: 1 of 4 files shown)


ptormene:
I guess dates (yyyy) should be checked to be four digits (unless we want to cope with very old stuff)

ptormene:
Probably, if we have the option to pick an "exact" date, and we choose to insert a range, the min and max should not be equal

ptormene:
Should the ground slope be limited to some degrees?

openquakeplatform/openquakeplatform/taxtweb/static/taxtweb/js/taxtweb.js
```diff
@@ -2533,16 +2559,185 @@ function taxt_BuildTaxonomy()
...
+    var height2 = $('#HeightCB2').val();
+    var height3 = $('#HeightCB3').val();
+    var height4 = $('#HeightCB4').val();
+    var date1 = $('#DateCB1').val();
+    var validated = false;
+    var validate_msg = "";
+    var h11 = true, h12 = true, h21 = true, h22 = true, h31 = true, h32 = true;
+    var d1 = true, d2 = true;
+
+    if (height1 > 0) {
```
ptormene: Please add comment like `known height`

ptormene:
When the permalink contains an invalid value, the corresponding field is not editable until you change the selection of its dropdown menu

ptormene:
LGTM

PR #430, opened 2015-06-15
Irv scale composite indicators

bwyss:
This PR includes provides the IRV web application with logic to check that the composite indicators (previously known as the primary indicators) are scaled, and if they are not, provide the user with a warning message and a button that will scale them.

To test this PR, simply load a project that does not have scaled primary indicators, for example the project called 'testing iri small UX improvements'. 
Once the project has loaded, select the primary indicator tab and you should see the warning message like this:

screen shot 2015-06-15 at 4 40 25 pm

After clicking on the button to scale the composite indicators, the composite indicators are scaled, and all the charts are re-rendered.

This PR also renames the primary indicators to composite indicators, per the project owners request.

### DIFF
--- a/openquakeplatform/openquakeplatform/static/js/irv_viewer.js
+++ b/openquakeplatform/openquakeplatform/static/js/irv_viewer.js
@@ -17,6 +17,7 @@
 
 $(document).ready(function() {
     $('#cover').remove();
+    $('.alert-unscaled-data').hide();
 });
 
 var layerAttributes;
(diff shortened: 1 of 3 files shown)


openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -35,6 +36,47 @@ var baseMapUrl = new L.TileLayer('http://otile1.mqcdn.com/tiles/1.0.0/map/{z}/{x
 var app = new OQLeaflet.OQLeafletApp(baseMapUrl);
 var indicatorChildrenKey = [];
 
+function scaleTheData() {
+    // Create a list of primary indicators that need to be scaled
+    // We are not scalling any of the IR indicators
```
ptormene: This is because we are not displaying any chart for the risk indicators, right?

openquakeplatform/openquakeplatform/static/js/irv_viewer_PCP_primary.js
```diff
@@ -74,22 +74,27 @@ function Primary_PCP_Chart(projectDef, layerAttributes, selectedRegion) {
...
         }
 
+        // Poipulate the object created above with values
         for (var n = 0; n < data.length; n++) {
             for (var o = 0; o < layerAttributes.features.length; o++) {
                 var field = data[n].plotElement;
                 var value = layerAttributes.features[o].properties[field];
+
+                if (value === null) {
+                    value = 211.309;
```
ptormene: I don't understand this

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: Yes, thats correct

openquakeplatform/openquakeplatform/static/js/irv_viewer_PCP_primary.js
bwyss: Whoops, thats not supposed to be there, it was just a test. But, yes, this gets addressed in PR #431

ptormene:
LGTM

PR #431, opened 2015-06-16
Irv manage incomplete data2

bwyss:
Per the request of the project owner, the charts will no longer be rendered when the composite data is incomplete. We now check for incomplete data, provide a warning message to the user and then stop the function from rendering the charts.

To test this PR, load a project that has incomplete data, you will then seem in each of the chart tabs a warning message like the this: 

screen shot 2015-06-16 at 12 22 50 pm

Please note that the tree chart is still visible and the user can still modify weights and save new versions of the project definition. The thematic map is not rendered to the map as the values are also dependent on the composite data.


ptormene:
LGTM

PR #432, opened 2015-06-19
Place the thematic map variable into the global scope

bwyss:
With the thematic in global scope, the application is able to remove any previously rendered map on each iteration of the block of code. This will prevent the maps overlapping eachother.

You can test this by modifying the web application so that it will render a new map (change weights, inversions, or make a new selection from the map indicator pull down menu. 

You will see that the map layers are not overlapping because the opacity of the map will remain constant (if the maps overlap, then the opacity will appear to change).


ptormene:
LGTM

PR #433, opened 2015-06-22
Update the GeoNode layers template

bwyss:
Update the GeoNode layers template in order to not show supplemental information when the category is SVIR

I needed to use stringformat in the filter because the category variable in surrounded by whitespace and a simple filter would not work.

To test, in the OQP open a layer the contains the category 'SVIR', you should NOT see the supplemental information metadata field. Then selected another layer that has category 'Hazard' (for example), and you should see the supplemental information metadata field


ptormene:
LGTM

PR #434, opened 2015-06-23
Check the svir plugin version

bwyss:
Check the svir plugin version, warn the user and stop the application if the version is older than 1.4.3

To test this feature, load a project that has a svir_plugin_version < 1.4.3. 
The application will render a warning message like this:

screen shot 2015-06-23 at 12 56 51 pm

If a project is loaded into the application with a version => 1.4.3, then the project will be loaded as normal.


ptormene:
LGTM

PR #436, opened 2015-06-29
Exposure export bug 1469636

pslh:
Fix for launchpad bug 1469636 
 https://bugs.launchpad.net/oq-platform/+bug/1469636

The key issue is that the study_id is passed to export API calls instead of the study_region_id.
This bug was not immediately obvious for two reasons:
*  for many studies (PAGER, most UN Habitat Level 0), study_id==study_region_id
*  the handling for sub-national studies is slightly different

As described in the launchpad bug, all NERA national studies are affected, as as the UN Habitat L0 studies for Mexico, Mongolia, South Africa and Ukraine. 

This pull request replaces study_id with the study_region_id from the selected study region.

In order to confirm that the coordinates are correct, I recommend loading the CSV data into QGIS over an OpenLayers background so as to check that the downloaded area corresponds to the expected region.  

This pull request also resolves two minor presentation problems:
* replaces "Has Residential" label with "Has **Non** Residential"
* resolves 404 error loading spinner image caused by use of Django verbatim directive

### DIFF
--- a/openquakeplatform/openquakeplatform/exposure/static/exposure/exposure_controller.js
+++ b/openquakeplatform/openquakeplatform/exposure/static/exposure/exposure_controller.js
@@ -177,7 +177,6 @@ app.controller('ExposureCountryList', function($scope, $filter, myService, ngTab
                 // Check the grid count
                 if ( $scope.selectedRegion[0].tot_grid_count < 300000) {
                     $('#exposure-building-form').append(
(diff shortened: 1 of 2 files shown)


daniviga:
LGTM

ptormene:
LGTM too

daniviga:
Should we push it on production as an hotfix? I would do it since it's just static content and it does not require a new deploy.

pslh:
@daniviga: Yes, I agree we should push to production.

daniviga:
@pslh: done, pushed to production

PR #440, opened 2015-07-21
Nested supplemental information

ptormene:
Addressing https://bugs.launchpad.net/oq-svir/+bug/1463720 (companion of https://github.com/gem/oq-svir-qgis/pull/83).
Changing the data structure of the supplemental information field, in order to keep separated the project definitions from other variables that are defined for the layer and that are shared by all the project definitions (e.g., the bounding box or the id of the layer on the Platform, that otherwise needed to be replicated for each project definition)


ptormene:
When adding a new project definition, the selected_project_definition_idx should point to the new one

openquakeplatform/openquakeplatform/svir/views.py
```diff
@@ -72,6 +72,48 @@
 @condition(etag_func=None)
 @allowed_methods(('GET', ))
 @sign_in_required
+def get_supplemental_information(request):
+    """
+    Given a layer name, read the corresponding supplemental_information from
+    the layer's metadata and return that as a json
```
bwyss: maybe this should read "Given a layer name, read the corresponding supplemental_information from the layer's metadata and return it as a json"

ptormene: Done

daniviga:
LGTM

PR #447, opened 2015-08-04
Irv conflicting labels

bwyss:
This PR provides a number of small UI/UX improvements to the IRV d3js tree chart.
Summary of fixes:
The project definition dialog loading spinner is now hidden only after the tree chart in rendered. This needed when the composite indicator data is incomplete. 

Resolved the labeling conflict between IR and SVI (when SVI has many children)

Resolved labeling issue when attribute label are more then 20 characters. In the case that attribute labels are more then 20 characters, I am substituting the attribute name with an acronym. This acronym can be hovered with the mouse, in which case the full attribute name is rendered on the top left of the dialog. Also note that all labels that can be hovered are rendered in blue, which gives the user a visual clue that it can be interacted with.

Result:
screen shot 2015-08-04 at 2 42 27 pm

### DIFF
--- a/openquakeplatform/openquakeplatform/static/js/irv_viewer_d3_tree.js
+++ b/openquakeplatform/openquakeplatform/static/js/irv_viewer_d3_tree.js
@@ -458,6 +458,7 @@
                 })
                 .attr("text-anchor", function(d) { return "end"; })
                 .text(function(d) {
+                    // Convert long attribute names text into acronyms
                     if (d.isInverted && d.name.length <= 20) {
                         return "- " + d.name;
                     } else if (d.isInverted && d.name.length > 20) {
--- a/openquakeplatform/openquakeplatform/static/js/irv_viewer_d3_tree.js
+++ b/openquakeplatform/openquakeplatform/static/js/irv_viewer_d3_tree.js
@@ -554,10 +554,12 @@
                 .style("fill", "#0000EE")
                 .attr("x", function(d) {
                     if ( d.field == 'SVI') {
-                        if (getRadius(d) > 15) {
+                        if (getRadius(d) >= 15 && getRadius(d) < 20 ) {
                             return "-4em";
+                        } else if (getRadius(d) >= 20) {
+                            return "-5em";
                         } else {
-                            return "-3  em";
+                            return "-2.7em";
                         };
                     } else{
                         return "-1em";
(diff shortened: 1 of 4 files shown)


openquakeplatform/openquakeplatform/static/js/irv_viewer_d3_tree.js
```diff
@@ -451,15 +451,50 @@
...
+                    } else if (d.isInverted && d.name.length > 20) {
+                        var matches = d.name.match(/\b(\w)/g);
+                        var acronym = matches.join('').toUpperCase();
+                        return "- " + acronym;
+                    } else if (d.name.length <= 20) {
                         return d.name;
+                    } else if (d.name.length > 20) {
+                        var matches = d.name.match(/\b(\w)/g);
+                        var acronym = matches.join('').toUpperCase();
+                        return acronym;
```
ptormene: It would be good to avoid duplicating the construction of `matches` and `acronym`. You could re-arrange conditions differently. So, if name length exceeds 20 characters, the acronym is built, and it is returned with a - if d.isInverted, or without the minus otherwise.

openquakeplatform/openquakeplatform/static/js/irv_viewer_d3_tree.js
```diff
@@ -511,9 +546,23 @@
             nodeEnter.append("text")
                 .attr("class", "pointer")
                 .style("fill", "#0000EE")
-                .attr("x", function(d) { return "-1em"; })
+                .attr("x", function(d) {
+                    if ( d.field == 'SVI') {
```
ptormene: you should do here as in the other PR, using the type instead of the name or field. Otherwise you might have that d.field is something like 'SVI_2' and it wouldn't work.

openquakeplatform/openquakeplatform/static/js/irv_viewer_d3_tree.js
```diff
@@ -511,9 +546,23 @@
...
+                        } else {
+                            return "-2.7em";
+                        };
+                    } else{
+                        return "-1em";
+                    }
+                })
                 .attr("dy", function(d) {
-                    if (typeof d.parent != "undefined" && d.x > d.parent.x){
+                    if (typeof d.parent != "undefined" && d.x > d.parent.x && d.field == 'SVI'){
```
ptormene: Same as above

ptormene:
LGTM

PR #448, opened 2015-08-05
Dynamic leaflet supplemental layers

bwyss:
This PR provides a dynamic dropdown menu featuring supplemental layers. The menu can be added to any leaflet based page. 

The way it works:
The JS makes an API call to Geoserver and then filters the list of all available layers that have a keyword "GEM-supplemental-layer". It then dynamically creates a dropdown menu with the layers.

You can see the menu in action, by opening the HRDE application. In the tool bar menu you can see the menu:
 
screen shot 2015-08-05 at 5 28 17 pm


daniviga:
LGTM as soon as #450 and #449 are merged

daniviga:
LGTM

PR #449, opened 2015-08-07
Add "GEM-supplemental-layer" keyword to ISC and GHEC

daniviga:


daniviga:
I've just found an issue. Closing for now.

daniviga:
The issue (duplicated item in "Download metadata") is not related to the fixture. Reopening.

daniviga:
For some reason commit 01d6d1405056f87767a9786157129b129152d509 is ignored by GS. Also other keywords that have been set previously in the ISC layer are not loaded when importing the layer. Probably we have a bug in the GS data importer.

PR #452, opened 2015-08-17
Fix the name of the EDUEACBGP indicator

ptormene:
There was a mismatch between the name of the indicator, incorrect, and its description, correct.
I could have changed the csv with the original data instead of writing a migration, but then we would have needed to re-import the whole db on all the working machines. Together with Matteo, we thought that implementing a migration would have been a cleaner approach.


daniviga:
Usually we put migrations in https://github.com/gem/oq-platform/tree/master/openquakeplatform/migrations since the migration that place can be run by the installation script. There's a reason why you have put it in a custom path? Maybe it must run only in special cases?

daniviga:
If the migration applies only to `dev_data` a custom path is fine, but I would move it into `dev_data/migrations`, since it is not strictly a `bin` command. What's your opinion?

ptormene:
I've moved the migration as suggested. We could merge it, and then open a separate PR to clarify that the data in dev_data is actually used in production, and to modify the installer accordingly.

daniviga:
LGTM

daniviga:
After some more investigations it seems the the data is actually loaded by the `deploy.sh`:
```bash
vir_dataloader () {
    local oqpdir="$1" db_name="$2" bdir

    if [ -f "private_data/svir.json.bz2" ]; then
        bdir="private_data"
    else
        bdir="${oqpdir}/svir/dev_data"
    fi
    openquakeplatform loaddata "${bdir}/svir.json.bz2"
}
```
https://github.com/gem/oq-platform/blob/master/openquakeplatform/bin/deploy.sh#L186

So the migration should be moved to the main `migration` folder: the `deploy.sh` will run it during installation. Please give the migration a proper incremental number.

daniviga:
I'm running the production installation test here: https://ci.openquake.org/job/zdevel_oq-platform/83/

daniviga:
I'm running the production installation test here: https://ci.openquake.org/job/zdevel_oq-platform/85/

daniviga:
Migration works and LGTM, but before merging please update also the `dev_data` to be consistent with the migration.

daniviga:
LGTM

PR #454, opened 2015-08-20
Risk Input Preparation Toolkit

bwyss:
**This Pull Request**

This PR is provides the OQP with a toolkit to prepare risk exposure, fragility and vulnerability NRML input models. The toolkit is organized into a single page with tabs for each category: exposure, fragility and vulnerability.

**Access**
One can access the toolkit by navigating in the OQP to Calculate-> Modeling Tools-> Risk Input Preparation Toolkit or by YourPlatformRootURL/risk_input_preparation_toolkit. One can also access (for the moment) a demo of the toolkit at: http://platform-dev-bmw.openquake.org/risk_input_preparation_toolkit/

**Use**
To use the input toolkit, each input tool includes inline tooltips which offer an explanation for each step needed to generate a NRML file.

**Note*** The toolkit is still a work in progress as the NRML structure is still in flux.

**Still to be done:**

 - Test the fragility and vulnerability with the reader for the new   
   schemas    
 - Provide the toolkit as a part of the virtual box images that are     
   distributed by GEM
 - Create a desktop application of the toolkit.      This    can be done
   using something like http://appjs.com/ or         
   https://github.com/atom/electron


openquakeplatform/openquakeplatform/static/css/risk_input_model.css
```diff
@@ -0,0 +1,103 @@
...
+	padding-right: 2%;
+	margin-left: 5px;
+}
+
+.ffsForm {
+	float: left;
+	margin-top: 20px;
+}
+
+.vfsForm {
```
ptormene: ffsForm and vfsForm might be unified, if their parameters remain identical

openquakeplatform/openquakeplatform/static/css/risk_input_model.css
```diff
@@ -0,0 +1,103 @@
...
+	width: 65%;
+	margin-left: 3px;
+}
+
+.ffsTableDiv {
+	padding-right: 40px;
+}
+
+.vfsTableDiv {
+	padding-right: 40px;
```
ptormene: Same for ffsTableDiv and vfsTableDiv

openquakeplatform/openquakeplatform/static/css/risk_input_model.css
```diff
@@ -0,0 +1,103 @@
...
+    padding-bottom: 8%;
+}
+
+.ffsTitle {
+    padding-left: 10%;
+    font-size: 18px;
+}
+
+.vfsTitle {
+    padding-left: 10%;
```
ptormene: Same as above

openquakeplatform/openquakeplatform/static/css/risk_input_model.css
ptormene: (like in http://stackoverflow.com/questions/2099252/2-classes-with-different-name-should-have-same-property-in-css-without-repeating)

openquakeplatform/openquakeplatform/static/css/risk_input_model.css
```diff
@@ -0,0 +1,103 @@
...
+	width: 600px;
+	height: 700px;
+}
+
+#economicCheckBoxes {
+	margin-left: 35%;
+}
+
+#occupantsCheckBoxes {
+	margin-left: 35%;
```
ptormene: same as above

openquakeplatform/openquakeplatform/static/js/risk_input_model_exposure.js
```diff
@@ -0,0 +1,437 @@
...
+});
+
+$('#defineCostBusiness').change(function() {
+    var defineCost = $('#defineCostBusiness').val();
+    if (defineCost == 'per_area') {
+        $('#perArea').show();
+        showArea = true;
+    }
+});
+
```
ptormene: Perhaps you could define a function similar to:
```javascript
function defineCost(costType) {
    var defineCost = $(costType).val();
    if (defineCost == 'per_area') {
        $('#perArea').show();
        showArea = true;
}
```
And call it in the 4 above cases

openquakeplatform/openquakeplatform/static/js/risk_input_model_exposure.js
```diff
@@ -0,0 +1,437 @@
...
+    }
+
+    // Default columns
+    header = ['id', 'longitude', 'latitude', 'taxonomy', 'number'];
+
+    function checkForValue (argument, valueArg) {
+        // Modify the table header only when the menu is altered
+        // This constraint will allow Limit, Deductible and Occupant elements to be
+        // added to the header
+        if (argument != 'none' && valueArg === undefined) {
```
ptormene: Please be sure this check works (see http://stackoverflow.com/questions/28583578/best-practice-of-checking-if-variable-is-undefined). For the reason explained in the top rated answer, I prefer using `typeof valueArg == 'undefined'`, which is safer in general because it avoids the possible reference error when `valueArg` was not declared with `var`. Please check also all the other occurrences of this.

openquakeplatform/openquakeplatform/static/js/risk_input_model_exposure.js
```diff
@@ -0,0 +1,437 @@
...
+    var nightInx = checkHeaderMatch(night);
+    var transitInx = checkHeaderMatch(transit);
+    var retrofittingInx = checkHeaderMatch('retrofitting');
+    var limitInx = checkHeaderMatch('limit');
+    var deductibleInx = checkHeaderMatch('deductible');
+    var assetIdInx = checkHeaderMatch(assetId);
+
+    // Create the asset
+    for (var i = 0; i < data.length; i++) {
+        var costTypes = '\t\t\t<costTypes> \n';
```
ptormene: Is the space after `>` desired, or is it just a typo?

openquakeplatform/openquakeplatform/static/js/risk_input_model_exposure.js
```diff
@@ -0,0 +1,437 @@
...
+        asset +=
+            '\t\t\t<asset id="'+id+'" '+number+' '+area+' '+taxonomy+' > \n' +
+                '\t\t\t\t<location '+longitude+' '+latitude+' />\n' +
+                costs +
+                occupancies +
+            '\t\t\t</asset>\n';
+    }
+
+    // Create a NRML element
+    var NRML =
```
ptormene: I guess there must be a simpler way in javascript to produce the xml, given tags and attributes

openquakeplatform/openquakeplatform/static/js/risk_input_model_fragility.js
```diff
@@ -0,0 +1,298 @@
...
+    // Logic to remove a table
+    $('.destroyTable').click(function() {
+        $('#table'+this.id).remove();
+        var removedTable = this.id;
+        delete activeTablesObj[removedTable];
+    });
+
+    // Increase the ffs panel when many limit states are defined
+    if (limitStateLength > 5) {
+        $('.ffsTableDiv').height(240 + (limitStateLength * limitStateLength));
```
ptormene: Why limitStateLength^2? It sounds like it's a little too much. Or I could be missing something

openquakeplatform/openquakeplatform/static/js/risk_input_model_exposure.js
bwyss: I see the advantage as explained in the stack exchange, but in this case it would be impossible for  argument or valueArg to be not declared as it is passed through the function. In this case I consider a strict comparison appropriate.

ptormene: Agreed

openquakeplatform/openquakeplatform/static/js/risk_input_model_exposure.js
bwyss: I am not very convinced that there is an issue with this approach? The XML that is created is very dynamic and thus I would be afraid that using a third party tool would only complicate the process.

openquakeplatform/openquakeplatform/static/js/risk_input_model_fragility.js
bwyss: Yes, 1.5 is probably a better fit

openquakeplatform/openquakeplatform/static/js/risk_input_model_exposure.js
ptormene: I would have probably explored the other approach (to me, it looked more promising), but I guess now that you have already produced this solution, it is probably better to keep it as it is.

openquakeplatform/openquakeplatform/static/js/risk_input_model_fragility.js
```diff
@@ -0,0 +1,300 @@
...
+    $('.destroyTable').click(function() {
+        $('#table'+this.id).remove();
+        var removedTable = this.id;
+        delete activeTablesObj[removedTable];
+    });
+
+    // Increase the ffs panel when many limit states are defined
+    if (limitStateLength > 5) {
+        console.log('limitStateLength:');
+        console.log(limitStateLength);
```
ptormene: Please remove console.log

openquakeplatform/openquakeplatform/static/js/risk_input_model_fragility.js
```diff
@@ -0,0 +1,300 @@
...
+        $('.ffsTableDiv').height(240 + (limitStateLength * limitStateLength));
```
ptormene: I am not sure I am reading this correctly, but it looks like you should set both the height and width, instead of setting the height to the square of limitStateLength.

openquakeplatform/openquakeplatform/static/js/risk_input_model_exposure.js
```diff
@@ -0,0 +1,428 @@
...
+$('#retrofittingSelect').hide();
+
+/////////////////////////////////////////////////////////
+// Manage the visibility of the perArea selection menu //
+/////////////////////////////////////////////////////////
+$('#perArea').hide();
+
+$('#defineCostStruc').change(function() {
+    defineCost($(this).val());
+    if (defineCost != 'none') {
```
ptormene: It looks like defineCost is undefined, and you have to use $(this).val() instead

bwyss: It is because I used the same name as the function

ptormene: I am not sure it works though. Please double-check it.

bwyss: Seems to be working

ptormene: If you select Structural Costs different than "None", the "Retrofitting Cost" checkbox is shown. But if afterwards you select "None" again, the checkbox doesn't disappear. I believe substituting `defineCost` with `$(this).val()` at line 36, it will work correctly.

micheles:
LGTM

ptormene:
LGTM

PR #455, opened 2015-08-20
Add API to get all the general_information instances, serialized in a json format

ptormene:
To use the new API, you can call .../vulnerability/general_information_json_get_all
It should return a json list with the serialization of all the instances of general_information that are editable by the user who is currently logged in.

This PR also fixes a small issue in how json objects are passed to the html pages. Before, they were passed as strings, and they required to be parsed client-side. Now the parsing is not necessary anymore, because the object is recognized as a json.

NOTE: For dozens of curves, the whole serialization is fast (about 1 second) but for thousands it might become time- and memory-consuming, so we might consider different approaches, serializing just a subset of the available curves. This shouldn't be a problem though, because this API is supposed to be used only for testing purposes, on a reasonable quantity of test data and not on huge production data.


openquakeplatform/openquakeplatform/vulnerability/urls.py
```diff
@@ -40,6 +40,7 @@
     url(r'^view/(?P<gen_info_id>\d+)(/(?P<action>[a-zA-Z_-]+)?)?$',
         views.show_general_information, name='view'),
     url(r'^data/(\d+)/$', views.general_information_json_get),
+    url(r'^data$', views.general_information_json_get_all),
```
daniviga: Please move this line above the `/data/id` match

daniviga: Consider to change `r'^data$'` into `r'^data/?$'` which will match also with the trailing slash. This should work with the right orders of the matches (see comment above)

daniviga:
LGTM

PR #456, opened 2015-08-25
Fix the majority of pep8 issues in the Vulnerability app

ptormene:
I did not remove some unused variables, because I suppose they might become useful in the future, but I've added comments in those cases. There are also some long lines that would become a little harder to read if they are split into separate lines, so I kept them as they are.


daniviga:
LGTM

PR #459, opened 2015-09-04
svir: drop theme_id column from indicator and derive it as a property

ptormene:
Otherwise there was some redundancy, and it would have been possible to link an indicator to a theme and at the same time to a sub-theme that does not belong to the same theme. The same structure will also be used for the sub-national DB.
Since the indicator's theme is now derived as a property, this change does not affect the APIs and I've tested that things keep working when loading data from the platform to the QGIS plugin.


ptormene:
WIP, because I forgot to modify the fixtures

ptormene:
Dev data fixtures are now modified. Production fixtures are changed in the companion https://github.com/gem/oq-platform-data/pull/3

daniviga:
LGTM

PR #463, opened 2015-09-15
Add support for heatmaps in Icebox hazard

daniviga:
WIP. Will be part of 2015-09 sprint.


daniviga:
Closing since Icebox will probably die soon

PR #473, opened 2015-09-25
Exposure: add num_l2_studies to the output of get_all_studies API

ptormene:
Please note that this changes the output of the API. In particular, `num_l1_studies` is not returned anymore. Instead, `num_studies`, `num_l1_names` and `num_l2_names` are returned (please see the new docstring for further details)


ptormene:
We might need to add a ORDER BY to the external query. For instance, by country_name.

bwyss:
The 'order by' seems to be fine in the client application

bwyss:
LGTM

ptormene:
Even though the client application orders the items correctly, I guess we should add a simple order by to the DB query, to improve the output of the API. I believe for Paul it would be a very quick task.

PR #474, opened 2015-09-28
Exposure ordering bug2.0

bwyss:
This PR began as investigation into two bugs outlined in the Launchpad bug report. However two addition bugs where found (see LP comments).

LP bug: https://bugs.launchpad.net/oq-platform/+bug/1484920


ptormene:
Unfortunately, I will not be able to review this for the next two weeks. If it is more urgent, I guess it would be better to assign this code review to somebody else.

openquakeplatform/openquakeplatform/exposure/static/exposure/exposure_controller.js
```diff
@@ -41,7 +41,7 @@ var downloadFractions = function() {
         $.ajax({
             type: 'get',
             url: 'export_fractions_by_study_region_id?sr_id='+sr_id,
-            success: function(data, textStatus, jqXHR) {
+            success: function(data, textStatus, jqXHR) {;
```
nastasi-oq: What is the meaning of `;` character  ?

openquakeplatform/openquakeplatform/exposure/static/exposure/exposure_controller.js
```diff
@@ -279,18 +313,18 @@ app.controller('ExposureRegionList', function($scope, $filter, $http, myService,
...
-                '<label id="id_residential_1_text" for="id_residential_1"><input class="exposure_export_widget" id="id_residential_1" name="sub-residential" type="radio" value="non-residential" /> Non-Residential</label></br>'+
-                '</p>'+
-                '<p><label for="id_outputType_0">Output Type:</label></br>'+
-                '<label for="id_outputType_0"><input class="exposure_export_widget" id="id_outputType_0" name="sub-outputType" type="radio" checked="" value="csv" /> CSV</label></br>'+
-                '<label for="id_outputType_1"><input class="exposure_export_widget" id="id_outputType_1" name="sub-outputType" type="radio" value="nrml" /> NRML</label></br>'+
-                '</p>'+
-                '<input type="hidden" name="sub-study" value="'+study.study_region_id+'">'+
-                '<br>'+
+            '<p><b>Download Region Wide Building/Dwelling Fractions:</b></p><button id="dwellingFractionsDownload" type="button" value="'+study.study_region_id+'">Download</button></br></br>'+
+            '<b>Download Gridded Sub-National Building Exposure:</b></br>'+
```
nastasi-oq: Could we move all this HTML code to the html page/template and hide it with a `display: none` css directive ?
Before show it we can set the `sub-study` input tag with a proper jquery call.

openquakeplatform/openquakeplatform/exposure/static/exposure/exposure_controller.js
```diff
@@ -71,10 +71,25 @@ var showErrorDialog = function(message, options) {
     $("#error-dialog").dialog(options);
 };
 
-app.controller('ExposureCountryList', function($scope, $filter, myService, ngTableParams) {
+app.controller('ExposureCountryList', function($scope, $filter, myService, ngTableParams, $http) {
+
+    // Set up the templates
+    $scope.templates = [
+    {
```
nastasi-oq: Why, instead of use complex template approach use a more simple `ng-show` directive with different comparisons ?
In this way all the html code will be in the same place but just what you want to show will be visualized.
Here a simple example: http://jsfiddle.net/nastasi/g4mfz5c2/

nastasi-oq:
LGTM

PR #475, opened 2015-09-29
Angular empty selection bug

bwyss:
In the web applications that are using AngularJs for dynamic and responsive selection menus (IRV & HRDE), there is an annoying bug where each menu has an empty option as the first child of selection element. 

screen shot 2015-09-29 at 11 13 23 am

This PR includes a fix for this bug as found on the second highest rated answer here: http://stackoverflow.com/questions/12654631/why-does-angularjs-include-an-empty-option-in-select

With this solution the menu now does not include an empty first child element:

screen shot 2015-09-29 at 11 15 23 am


nastasi-oq:
LGTM

daniviga:
[Hotfix] This commits have been cherry-picked in the platform-1.2 branch for the 1.2.1 release.

PR #476, opened 2015-09-29
Update the IRV unit tests per the recent QGIS toolkit refactor

bwyss:
The QGIS toolkit has been updated to provide a layer tag "irmt_plugin_version" as apposed to the older "svir_plugin_version". This PR updates the IRV Jasmine unit tests to support both options.


daniviga:
I think this should be an hotfix for 1.2.1 since 1.2.1 is targeted for the plugin version 1.7.1 which has the `IRMT_PLUGIN_VERSION` variable name.

daniviga:
I see that the fix for the actual code is just a couple of `;` and the rest is for the tests. Anyway, let's it enter the 1.2.1 if you agree.

daniviga:
LGTM

daniviga:
[Hotfix] This commits have been cherry-picked in the platform-1.2 branch for the 1.2.1 release.

PR #480, opened 2015-10-02
Refactor the version comparison function

### LINKED ISSUE #479: The check on QGIS plugin version is too strict
We are currently checking all the three version numbers, major, minor and maintenance: version 1.7.1 of the plugin is not working because we are looking for `COMPATIBILITY_VERSION = 1.7.0`.

We should check up to the second number (minor) only, since in any case no changes in the protocol should be expected without increasing the minor number.

```
COMPATIBILITY_VERSION = 1.7.0
1.6.x -> KO
1.7.0 -> OK
1.7.1 -> OK
...
1.7.x -> OK
```
We should also decide what to do if the minor number is greater, like 1.8.0 is this fine or not if the COMPATIBILITY_VERSION is 1.7.0 ? (I vote NO for the current status, but it should be YES)

bwyss:
This update will:
1) check for a match between the major release numbers
2) check for match between the minor release numbers
3) not check for patch numbers

This means that any version number with 1.7.X will be compatible with
the IRV app.

Closes #479

### DIFF
--- a/openquakeplatform/openquakeplatform/static/js/irv_viewer.js
+++ b/openquakeplatform/openquakeplatform/static/js/irv_viewer.js
@@ -926,6 +926,7 @@ function getGeoServerLayers() {
 
 
 function versionCompare(a, b) {
+    // This version check ignores the patch numbers
     if (a === undefined) {
         return -1;
     }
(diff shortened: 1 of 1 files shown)


daniviga:
LGTM

PR #482, opened 2015-10-08
Implement Mapbox-GL into IRV App

bwyss:
This PR adds the Mapbox-Gl Js library. 

**Why:**
Rendering pure vector geometries into the web map is required by the IRV app because each time the SVIR indicator weights are modified, new values are calculated and thus thematically mapped.
Rendering pure vector geometries into the web map has obvious limitations and can not be scaled.
The solution it to use the Mapbox-Gl Js library and render vector webGL tiles into the web map. There is a longer initial load time as a result, but the app can not manage far larger geometries.

**How:**
The GeoJSON returned from GeoNode/Geoserver is passed directly to Mapbox-GL which uses it to create vector tiles. 

**Workflow**
With regards to the mapping features of the IRV app, the app now checks the browser for Web-GL support. If the browser does not support Web-GL then it will check the number of vertices. If the project is over the vertices threshold, then a error is provide to the use. If the project is under the vertices threshold, then the application will continue with the old leafletJS thematic mapping solution.
However, if the browser does support Web-GL, then Mapbox-gl is used for  thematic mapping.

**Notes**
The previously implemented leaflet thematic mapping approach is still implemented and is used as a fallback if the users browser does not support WebGL and if the vertices count is less then the predetermined threshold.
This implementation of Mapbox-GL does not require any additional server side dependancies.
Additional efficiency may be gained by improving the map source/style refreshing method.
Additional efficiency will be gained by moving away from GeoJSON provided by GeoServer to a vector tile server solution.


openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -775,13 +788,350 @@ function scale(IndicatorObj) {
...
+    // Focus map on layer
+    if (projectChange) {
+        map.fitBounds(mapboxBoundingBox);
+    }
+
+    var min = Math.min.apply(null, minMaxArray).toFixed(2);
+    var max = Math.max.apply(null, minMaxArray).toFixed(2);
+    min = (parseFloat(min) - 0.1);
+    // Round up the max
+    max = Math.ceil(max * 10) / 10;
```
ptormene: Why not just `max = Math.ceil(max)`?

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -775,13 +788,350 @@ function scale(IndicatorObj) {
...
+        // FIXME would really prefer to update the map source here
+        map.removeSource('projectSource');
+    }
+
+    // Case 4 destory the source
+    if (projectChange) {
+        try {
+            map.removeSource('projectSource');
+        } catch (e) {
+            // continue
```
ptormene: What kind of issues could we have here? Perhaps we should return some readable error if anything fails.

openquakeplatform/openquakeplatform/static/js/irv_viewer_d3_tree.js
```diff
@@ -387,6 +393,9 @@
                     }
                 }
             }
+            console.log('weightChangeLevel:');
+            console.log(weightChangeLevel);
```
ptormene: Please remove console logs

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: This values is used for the map filter scale breaks (L1035), which are being generated dynamically. So I need to round up. But if the max value is less then 1, for example 0.6555555, I do not want to round up to the nearest whole number (1), instead I want to round up to the first decimal place (0.7)

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: As far as I can tell there is no reason to manage the error. There are 2 cases: 
1. When a project is loaded for the first time. In this case there is no existing map source or layer, so we try to remove them, but as they do not exist, the app moves on and builds a new source and layer.
2 When a project is loaded for the 2nd, 3rd, nth time. In this case we find that there is a map source and layer that need to be removed, after being removed, the new map source and layers are created.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
ptormene: Ok

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
ptormene: Ok, perhaps you could add a short comment to the code, to make it more explicit.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -775,13 +788,350 @@ function scale(IndicatorObj) {
...
+        // 1) the level of the weight changes is lower than the level of the thematic map menu.
+            // Example 1: if the user is viewing the SVI thematic map, and then changes any theme, or
+            // primary indicator weight, the map must be refreshed.
+            // Example 2: If the user is viewing a theme thematic map (e.g. Economy), and then changes
+            // the SVI weight, the map should not be redrawn.
+        // 2) the thematic map menu has been changes
+        // 3) a new project is loaded into the map (case: weightChange = 0)
+    var selectedIndicatorLabel = $('#webGlThematicSelection option:selected').attr('class');
+
+    if (weightChange < parseInt(selectedIndicatorLabel || selectedIndicatorLabel !== '4')) {
```
ptormene: Please add a comment to clarify this line. It looks very cryptic to me. I don't understand `||` (a `&&` would look more intuitive, but we have to be sure about the logic). Also, please specify what does `weightChange` means and what kind of values it might take.

openquakeplatform/openquakeplatform/test/config.py
```diff
@@ -0,0 +1,5 @@
+pla_basepath="http://192.168.56.10:8000"
```
ptormene: Is this supposed to be a local file, or should it be tracked by git?

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: in JS || = or, while && =  and

ptormene: Yes, of course. My point was that it looks like a "or" is a bit strange in that context, whereas a "and" would made me less surprised. I suppose I am still not grasping the logic of that line.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: Done

ptormene:
Please open a ticket (or solve) the issue we've seen about setting a weight to 0

bwyss:
I submitted a bug into LP: https://bugs.launchpad.net/oq-platform/+bug/1508820

daniviga:
Hey @bwyss, please use the GitHub issues, we have deprecated launchpad (see the message on its homepage)

bwyss:
https://github.com/gem/oq-platform/issues/497

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -511,7 +535,7 @@ function processIndicators(layerAttributes, projectDef) {
     // SVI is an object with region and value
     if (svThemes) {
         SVI = combineIndicators(sviNameLookUp, themeData, sviJSONthemes );
-        scale(SVI);
+        //scale(SVI);
```
ptormene: Please check about this line (to see if you want to keep or delete it)

ptormene:
I've just posted this issue: https://github.com/gem/oq-platform/issues/504

ptormene:
In the web viewer we have the same behavior described in https://github.com/gem/oq-irmt-qgis/issues/100. We should decide how to proceed with respect to that. Anyway, this does not specifically affect this PR.

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -775,13 +799,397 @@ function scale(IndicatorObj) {
...
+                '<ul id="legendLables" class="legend-labels">'+
+                '</ul>'+
+            '</div>'+
+        '</div>'
+    );
+
+    // Cases 1, 2, 3, 4, and 5
+    //if (weightChange === false) {
+        // Create a new mapbox layers
+        for (var i = 0; i < 6; i++) {
```
ptormene: Please check about the commented code and see if possibly deleting it and re-indenting the inner block

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -775,13 +799,397 @@ function scale(IndicatorObj) {
...
+            map.addLayer({
+                'id': i,
+                'type': 'fill',
+                'source': 'projectSource',
+                "source-layer": "eq-simple",
+                'interactive': true,
+                //'text-field': '{Parroquia}',
```
ptormene: can this line be deleted?

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
```diff
@@ -775,13 +799,397 @@ function scale(IndicatorObj) {
...
+    // Add all primary indicators to selection menu
+    $('#leafletThematicSelection').append('<optgroup label="Composite Indicators">');
+    for (var ic = 0; ic < allPrimaryIndicators.length; ic++) {
+        $('#leafletThematicSelection').append('<option>'+allPrimaryIndicators[ic]+'</option>');
+    }
+
+    // set the map selection menu to IRI or previously selected indicator value
+    if (selectedIndicator == undefined) {
+        // TODO fix this to use optgroup:first
+        //$('#leafletThematicSelection').val('IRI');
```
ptormene: Please check about this comment

CHANGELOG
```diff
@@ -1,3 +1,8 @@
+
+  [Ben Wyss]
+  * Added the Mapbox-Gl Javascript library in order to render vector
+    tiles into the IRV application.
+
```
ptormene: If you think it's worth it, you might list here also other improvements provided by this PR

ptormene:
Besides minor comments, LGTM

openquakeplatform/openquakeplatform/static/js/irv_viewer.js
bwyss: there is already a ticket for this known issue: https://github.com/gem/oq-platform/issues/505

PR #489, opened 2015-10-20
Risk ipt missing loss category

bwyss:
This Pr fixes 2 small bugs in the Risk input preparation toolkit.
The first bug is that fragility functions did not include the loss category.
The second bug is that there was extra white space in front of the limit states tag for fragility / continuous functions.

### DIFF
--- a/openquakeplatform/openquakeplatform/static/js/risk_input_model_fragility.js
+++ b/openquakeplatform/openquakeplatform/static/js/risk_input_model_fragility.js
@@ -204,6 +204,7 @@ $('#saveBtnFF').click(function() {
 
     var functionId = $('#functionId').val();
     var assetCategory = $('#assetCategory').val();
+    var lossCategory = $('#lossCategory').val();
     var functionDescription = $('#functionDescription').val();
 
     /////////////////////////////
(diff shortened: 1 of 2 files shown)


daniviga:
LGTM for me, but let's wait for @CatalinaYepes LGTM

CatalinaYepes:
The bugs have been solved. However it is also required an additional check regarding the `Limit States` box:
Users can not add free text. Only letters, numbers and underscored are permitted in a single string per limite state (e.g. [no_damage, slight, collapse])

micheles:
Also notice that users could make case errors (i.e. NoDamage is different than nodamage). Perhaps the tool should produce a fixed case (say all lowercase). However this is something to discuss and not needed for the risk workshop.

CatalinaYepes:
This issue is also important in the Vulnerability Database because the export option should transform a free text into NRML format. @nastasi-oq

daniviga:
So, should I merge this or not?

daniviga:
Let's merge this one, more fixes should be addressed in a new PR

daniviga:
[Hotfix] These commits have been cherry-picked in the platform-1.2 branch for the 1.2.4 release.
[Hotfix] This fix has been manually pushed to production since release 1.2.4 is currently blocked.

PR #514, opened 2015-11-25
svir: fix filtering indicators by theme

ptormene:
Recently, to avoid potential inconsistencies, we changed the `theme` attribute of `Indicator` into a property, so the `Indicator` is directly linked to a `subtheme` and the `subtheme` belongs to one of the `themes`. But we forgot to change one filter, so the API was unable to retrieve indicators belonging to a given `theme` if no `subtheme` was specified. This PR fixes such case.


mbernasocchi:
LGTM

PR #516, opened 2015-11-27
Irv single primary indicator corner case

bwyss:
There was a small error in the way that the IRV application managed the corner case when a single primary indicator is provided in the project definition. Specifically, the issue was within the scaling function.

In addition, this PR resolves another small bug which provides the user with a more complete error message when the composite indicator data is incomplete.

With regards the the web app, the purpose of managing all of the indicator values is to be able to plot them in the various charts. However if a single primary indicator is provided, then by default the IRI chart is disabled because an IRI value can not be computed without IR and SVI

The theme chart also must be disabled as it is designed to identify differences between multiple themes.

And the Primary indicator chart, who's purpose is to identify outlining data between multiple indicators, so the chart has no purpose when provided with a single indicator.

For these reason, when a single values is provided, or when the values are all the same (eg. max == min) I am disabling all the charts. 
So, it does not make any difference what the final scaled value is, as it is never used by the web app


daniviga:
Test are running here https://ci.openquake.org/job/zdevel_oq-platform/182/

daniviga:
LGTM if tests are geen

PR #520, opened 2015-12-14
Disable the icebox application

daniviga:
Icebox isn't supported since OpenQuake Engine 1.6. Users can use OATS instead.

Test: https://ci.openquake.org/job/zdevel_oq-platform/185/


micheles:
LGTM if the tests are green

PR #528, opened 2016-01-14
mantain line color for a nation when adding more

mbernasocchi:
This PR allows keeping the same color for a line dedicated to a country in the PCP chart.
this is my first platform PR so be nice ;)


ptormene:
LGTM

PR #529, opened 2016-01-19
Ript: validate using nrml as a library

nastasi-oq:
Instead of use engine-server as service (with all the testing and development complexities inherited by this choice) oq-risklib is used as library in a new view method.

Tests pass: https://ci.openquake.org/job/zdevel_oq-platform/196/


nastasi-oq:
LGTM

PR #531, opened 2016-01-21
selenium: make screenshot only if Firefox driver is present

ptormene:
...in order to avoid hiding an error message caused by missing config


nastasi-oq:
LGTM

PR #536, opened 2016-01-27
IRV: make the d3 tree layout as it is in the QGIS plugin

ptormene:
Since in a previous PR we have enabled zooming and panning, now we can display the whole attribute names in the tree, instead of producing just an acronym when the name is long. Labels need to be placed a little differently, and nodes need to be more spatially separated, in order to avoid overlapping. The tooltip can do the same thing as in the QGIS plugin, displaying the attribute name as it is inside the layer (while the full name is shown in the tree).

Tests are green: https://ci.openquake.org/job/zdevel_oq-platform/202/


nastasi-oq:
LGTM

Issue #540, opened 2016-02-01
IRV - Adding new project definitions causes some wrong corner cases


ptormene:
1. In some cases, the check on the existence of the new project definition name fails, saying it already existed although it did not. It does not happen always, and probably it's due to the fact that some events can occur with different delays, causing different behavior depending on the order things happen.
2. When a new project definition is added, it is also pre-selected from the dropdown menu. If afterwards another one is selected, and then the new one is re-selected, it can not be loaded. The only way to be able to use it afterwards is reloading the whole page, which makes the new item actually available.
3. I've also noticed a redundant check on the existence of an assigned name for a new project definition. The check is already done by the API that is called by the IRV app, so there is no need to repeat the same check here.
4. In some cases, the addition of a new project definition is performed multiple times redundantly, instead of triggering it just when the submit button is clicked. This seems to be due to unnecessarily nested event definitions.
5. The code contains some unnecessary delays, that were added to partially fix some of the above issues, but that should be removed once the real causes of these issues are fixed.

PR #541, opened 2016-02-02
IRV: use api check for unique name

ptormene:
Addressing points 1 and 3 of #540, removing some unused code and fixing some minor details that caused warnings by my syntax checker.

The `addProjectDefinition` function was not used anywhere, and exactly the same code was already used elsewhere, so I kept only the other (used) block.

The main change made by this PR is to leverage the `svir/add_project_definition` API to check the uniqueness of the name assigned to the new project definition. The IRV app was doing the same check, which was redundant and has been therefore removed.


ptormene:
I have also rearranged some code that was unnecessarily nested.
Another small change avoids a recursion error, but it still does not fix the fact that newly added project definitions can not be correctly selected without reloading the whole page.

nastasi-oq:
LGTM

PR #544, opened 2016-02-02
IRV - Fix the selection of newly added project definitions

ptormene:
Final steps to fix https://github.com/gem/oq-platform/issues/540

The issue was partially fixed by https://github.com/gem/oq-platform/pull/541. After cleaning the project definition in order to avoid a recursion issue, we also had to re-parse it before adding it to the list of available project definitions.

I've also solved an annoying issue that was due to the fact that the handler for the submission of a new project definition was re-defined and triggered every time the `loadPD` function was called, instead of defining it separately and calling it only when the `submitPD` button is triggered. Extracting that function, the events work in the right order and we don't need anymore those ugly workarounds (timeouts, etc) that only partially solved the effects of this misbehavior. One of the removed timeouts was 3 seconds, which made the application slow even for small projects.

### DIFF
--- a/openquakeplatform/openquakeplatform/irv/static/irv/js/irv_viewer.js
+++ b/openquakeplatform/openquakeplatform/irv/static/irv/js/irv_viewer.js
@@ -33,7 +33,6 @@ var thematicLayer;
 var leafletBoundingBox = [];
 var mapboxBoundingBox = [];
 var license;
-var projectDefUpdated;
 var webGl;
 var projectChange = false;
 var verticesCount = 0;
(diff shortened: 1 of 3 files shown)


openquakeplatform/openquakeplatform/irv/static/irv/js/irv_viewer.js
```diff
@@ -1258,27 +1257,21 @@ function thematicMapCreation() {
     $('#absoluteSpinner').hide();
 }
 
-function watchForPdSelection() {
+function whenProjDefSelected() {
+    $('#pdSelection').prop("disabled", true);
```
nastasi-oq: The '#pdSelection' probably must be disabled inside the fuction fired by '#submitPd'  click and reenabled after .post async call returned.

openquakeplatform/openquakeplatform/irv/static/irv/js/irv_viewer_d3_tree.js
```diff
@@ -696,4 +638,53 @@
         $('#projectDef-spinner').hide();
     } //end d3 tree
```
nastasi-oq: Please change this comment with the real name of the closed function.

openquakeplatform/openquakeplatform/irv/static/irv/js/irv_viewer.js
ptormene: Done. I'm keeping it also here, because the updating process triggered by the selection might take time, and we should not enable selecting a different project definition until such operation has been completed.

openquakeplatform/openquakeplatform/irv/static/irv/js/irv_viewer_d3_tree.js
ptormene: Done

nastasi-oq:
LGTM

PR #546, opened 2016-02-04
Tests timeouts fix

nastasi-oq:
Seems that a transient inconsistency in selenium is reflected in our tests broken them.
Added a delay between 2 critical operation is a good workaround.
Test are running here: https://ci.openquake.org/job/zzdevel_oq-platform/
and we scheduled a bunch of new jobs instances to be sure the problem is constantly solved.


daniviga:
LGTM

PR #549, opened 2016-02-11
IRV - fix corner case in which a zone with no available data is clicked on the map

ptormene:


nastasi-oq:
LGTM

ptormene:
I will update the changelog before merging

PR #552, opened 2016-02-16
IRV - Use separate widgets (instead of tabs), to display charts

ptormene:
While working in the direction described by the title, I made a couple of improvements/fixes:

* moving the mouse over the median line, now the text 'median' is correctly displayed for all charts (before, it produced errors in some cases)
* I've added the zooming/panning functionality to all the charts, also removing the scrollbars to those widgets and fixing the layouts in order to let the charts fill the whole available space 

Tests are green: https://ci.openquake.org/job/zdevel_oq-platform/233/


nastasi-oq:
LGTM

daniviga:
LGTM too

PR #554, opened 2016-02-19
Taxtweb header

nastasi-oq:
Splitting taxtweb in 3 different include files allow us to create 2 templates to include optionally platform headers and footer.


daniviga:
LGTM

PR #562, opened 2016-02-29
ript: hotfix for Aggregated Structural Costs case syntax error.

### LINKED ISSUE #561: RIPT - Exposure - extra " character in the output file.
When generating an exposure model using the aggregated value option, and extra “ is written in the output file. I used the attached exposure information (exposure_example.txt) which led to the exposure_example.xml also attached.

nastasi-oq:
When generating an exposure model using the aggregated value option, and extra “ is written in the output file. I used the attached exposure information (exposure_example.txt) which led to the exposure_example.xml also attached.

It fixes #561.


daniviga:
LGTM

Issue #504, opened 2015-11-04
IRV - Changing the color theme many times results in an increasingly slow response


ptormene:
The `webGlColorPicker` menu seems to work at first, but on my computer (both using firefox and chrome) if I switch to different colors many times, the response time approximately doubles at each attempt. It is very quick at first but, after about 6 consecutive changes, the response becomes quite slow and it causes my firefox to open the following warning popup:
```
A script on this page may be busy, or it may have stopped responding. You can stop the script now, open the script in the debugger, or let the script continue.

Script: https://api.tiles.mapbox.com/m…box-gl-js/v0.11.1/mapbox-gl.js:194
```
I guess we should also decide if the possibility to change colors is actually a useful feature.

PR #564, opened 2016-03-02
IRV - Remove color themes

ptormene:
I believe color themes are not actually useful in this application, because:

1. We display one single layer at a time, so we don't need to differentiate styles of different layers
2. Below the layer, we always have OpenStreetMap (and the user can't change it), so we just need to use colors that look fine with it.

Furthermore, the theme management contained a bug, that caused the application to become slower and slower every time a different color theme was selected. Instead of attempting to fix the bug, I am rather proposing to remove the unused feature (resolves https://github.com/gem/oq-platform/issues/504)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
   [Paolo Tormene]
   * irv: less redundant calculations and better management of operators
+  * irv: remove color themes (feature that was not needed and that contained a bug)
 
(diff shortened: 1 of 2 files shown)


ptormene:
Chris already gave his approval to this change

nastasi-oq:
LGTM

PR #566, opened 2016-03-11
HRDE: open terms of use in a new tab instead of the current one

ptormene:
Otherwise clicking the button would cause to discard whatever you had already done in the hrde.


daniviga:
LGTM!

PR #568, opened 2016-03-15
IRV: make documentation with sphinx and unify the cases with/without webgl versions

ptormene:
This PR makes a copy of http://www.globalquakemodel.org/openquake/support/documentation/platform/irv/
and
http://www.globalquakemodel.org/openquake/support/documentation/platform/IRI-no-webgl/
using sphinx, and it unifies the cases with/without WebGL, that were almost identical. I haven't changed anything else, apart from those few things needed to unify the two cases. As a first step, we could keep it as it is, although it is a bit short and not completely up to date. Improvements and updates could be done in a separate PR.

I've created this documentation inside the `irv` app, with the idea that each app in the platform will have its own documentation. We might consider creating also a more general container, describing the whole platform and the features that many apps have in common.

When we merge this PR, we must remember to dismiss the old documentation and to properly link the new one instead.

We need to check if the version of sphinx we use while packaging works correctly, especially with the :numref: feature (numbered figures).
Before merging, I suppose we should automatize the compilation process for the documentation.


daniviga:
Please move to oq-platform2 if still applicable

Issue #571, opened 2016-03-30
IRV - Handling layers with incomplete data


ptormene:
Currently, if there is some missing data in the layer, the IRV application doesn't show anything in the map, and neither in the charts. It only shows the following message: `The application is not able to render charts or the map layer for this project because the composite indicator data is incomplete.`. Furthermore, such message appears when the project is loaded, but it disappears afterwards, for instance when a widget is hidden and shown again.
We definitely need to handle missing data properly, showing meaningful results for all the features for which primary and composite indicators can be calculated.

PR #573, opened 2016-03-31
IRV: handle layers with incomplete data

ptormene:
Fixes https://github.com/gem/oq-platform/issues/571

Tests are green: https://ci.openquake.org/job/zdevel_oq-platform/245/

I've managed to display charts also in case of some missing data, but currently null values are displayed as 0. Instead of that, we should remove lines connected to that missing point or make it more evident to the user that that number is not a zero but a missing value.
Median lines are not computed correctly, because the sum is divided by the total number of values, instead of taking into account only non-null values.


ptormene:
I am replacing this with a new updated PR

Issue #567, opened 2016-03-11
HRDE - Once a supplemental layer is added, it can not be removed


ptormene:
The only way to remove supplemental layers, once they have been added to the map, is to reload the page.
This should be probably addressed together with https://github.com/gem/oq-platform/issues/525

Issue #525, opened 2016-01-13
HRDE - "Supplemental Layers" selection re-adds layers instead of just re-selecting them


ptormene:
When a layer is selected from the "Supplemental Layers" dropdown menu, it is added to the map. If the user then switches to a different layer, also that second one is added to the map, together with the first one. If afterwards the first one is re-selected, it is re-added to the map, instead of just re-selecting the existing one.
The fact that a simple dropdown selection enables to add multiple options to the map looks a little counter-intuitive. If we really want to enable multiple selections, we should probably do so by means of a mutliselect widget.

PR #577, opened 2016-04-11
HRDE: manage supplemental layers through the "Layer Manager" widget

ptormene:
Thanks to these changes, supplemental layers can be managed through the "Layer Manager" widget, together with the other layers. Therefore, once they have been added to the map, they can be shown/hidden through checkboxes and they can be moved towards the top or bottom of the stack of layers.

Fixes https://github.com/gem/oq-platform/issues/567
Fixes https://github.com/gem/oq-platform/issues/525 (the solution is still sub-optimal, but it is probably acceptable as it is)

Before this PR, supplemental layers were managed outside the hrde app, although they were used only by hrde. So I moved the corresponding code inside.

Tests are running: https://ci.openquake.org/job/zdevel_oq-platform/247/


nastasi-oq:
LGTM

PR #578, opened 2016-04-13
Renamed fab argument host to hostname because the first is reserved.

nastasi-oq:
Fabric us host as reserved variable to specify the name of target host to try to perform installation instead of localhost, hiding the argument 'host' of target function.


ptormene:
LGTM

PR #586, opened 2016-04-26
Remove Calc/Share/Exp from authentication

daniviga:
As requested by @pslh 
(and get rid of an old icebox setting)


nastasi-oq:
LGTM

PR #589, opened 2016-05-12
Get rid of icebox

daniviga:
Was already disabled, just remove all the code, tables and views.

Removing the icebox data in a production platform speeds up the `updatelayer` performance by a factor of ten.

Tests: https://ci.openquake.org/job/zdevel_oq-platform/278/


nastasi-oq:
LGTM

PR #595, opened 2016-05-27
IRV: add tests handling missing values

ptormene:
These additional tests check that the application keeps working as expected in case indicators contain any null values, as it should be after recent improvements.

In order to check it, I have duplicated some tests, slightly modifying the copies by replacing some values with some `null`.

Tests are green: https://ci.openquake.org/job/zdevel_oq-platform/299/


nastasi-oq:
LGTM

PR #598, opened 2016-06-21
Improve and update the calculate page

daniviga:
For the Engine 2.0

Now the links are pointing directly to the resources and they don't link to the website anymore (which is not well maintained).

I would suggest to improve more this page and remove pages for the website, making it just pointing here.


nastasi-oq:
LGTM

Issue #600, opened 2016-07-12
MapQuest basemaps are broken


daniviga:
See:

- http://devblog.mapquest.com/2016/06/15/modernization-of-mapquest-results-in-changes-to-open-tile-access/
- http://lists.osgeo.org/pipermail/geonode-devel/2016-July/001064.html and following

daniviga:
Closed by #601

PR #601, opened 2016-07-12
Remove MapQuest layers

daniviga:
See #600 
Tests: https://ci.openquake.org/job/zdevel_oq-platform/305/

This is a temporary fix: we should at least

- Add google (key required)
- ~~Add a new providers (like MapBox having a working key)~~ (supported only in GeoNode master, see http://lists.osgeo.org/pipermail/geonode-devel/2016-July/001075.html)


daniviga:
LGTM

PR #604, opened 2016-07-27
Ipt test to app

nastasi-oq:
IPT tests moved to oq-platform-ipt repository, modified all the infrastructure to support this new schema.
Tests are green: https://ci.openquake.org/job/zdevel_oq-platform/310/


openquakeplatform/openquakeplatform/test/satellites_test.py
```diff
@@ -0,0 +1,24 @@
+import unittest
+from inspect import isclass
+
+def import_by_name(clsname):
+    components = clsname.split('.')
+    leaf = components[-1]
+    mod = __import__(clsname, fromlist=[leaf])
+    return mod
+
```
micheles: Maybe instead of clsname it would be better to call it fullname or dotname

openquakeplatform/openquakeplatform/test/satellites_test.py
```diff
@@ -0,0 +1,24 @@
...
+def get_checks(clsname):
+    cls = import_by_name(clsname)
+    for objname in dir(cls):
+         obj = getattr(cls, objname)
+         if isclass(obj):
+             if issubclass(obj, unittest.TestCase):
```
micheles: Flat is better than nested and you can save an indentation level with

```
if isclass(obj) and issubclass(obj, unittest.TestCase):
   ...
```

micheles:
Apart from the minor comments LGTM

openquakeplatform/openquakeplatform/test/satellites_test.py
```diff
@@ -0,0 +1,24 @@
...
+                 newname = "%s__%s" % (clsname.replace('.', '__'), objname)
+                 globals()[newname] = obj
+                 obj.__name__ = newname
+
+for clsname in ['openquakeplatform_ipt.test', 'openquakeplatform_taxtweb.test']:
+    try:
```
micheles: Even here, clsname is actually the name of the module

PR #613, opened 2016-09-23
Vulnerability: export short taxonomies  in NRML files instead of full.

nastasi-oq:
When a curve or a set of curves is exported as NRML files GEM taxonomies were exported as full instead of short taxonomies; this PR fix the problem.
Concurrency between taxtweb tests and platform tests on splash window are managed.
Tests are green: https://ci.openquake.org/job/zdevel_oq-platform/315/


ptormene (approved):
Please add some comments where there are any magic numbers. Apart from this, LGTM

PR #614, opened 2016-09-26
vulnerability: 'taxonomy_text' field must be declared as not required

nastasi-oq:
Without this fix an empty `taxonomy_text` value generates an error message and not allows to save the curve.


ptormene (approved):
LGTM

PR #617, opened 2016-10-26
Pip engine

nastasi-oq:
Use oq-hazardlib/oq-engine from sources via `pip` instead of discontinued ubuntu precise packages.


daniviga (approved):
LGTM

PR #618, opened 2016-11-04
IRV - Custom operator

ptormene:
This PR enables the Integrated Risk Viewer (IRV) to visualize projects in which the Integrated Risk was obtained through a custom calculation. In this case, the IRI will not be recalculated, and its values will just be read from a field specified by the user. The project definition also contains a description of the customized calculation.

Please see the companion: https://github.com/gem/oq-irmt-qgis/pull/146

In my local machine tests are green.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,8 @@
 python-oq-platform (1.7.1)
+  
+  [Paolo Tormene]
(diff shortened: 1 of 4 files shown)


mbernasocchi:
lgtm

PR #620, opened 2016-11-22
Update doc

daniviga:
Development documentation refactoring.

A section about the standalone apps in production is still missing.

The final result can be viewed here: https://github.com/gem/oq-platform/blob/update-doc/docs/Development-environment-on-Ubuntu-12.04.md


docs/Development-environment-on-Ubuntu-12.04.md
```diff
@@ -6,32 +6,49 @@ These are the main steps to install the OpenQuake Platform on Ubuntu 12.04 LTS u
 sudo apt-get install build-essential python-dev python-mock python-imaging python-virtualenv git postgresql-9.1 postgresql-server-dev-9.1 postgresql-contrib-9.1 postgresql-9.1-postgis openjdk-6-jre libxml2 libxml2-dev libxslt1-dev libxslt1.1 libblas-dev liblapack-dev curl wget xmlstarlet imagemagick gfortran python-nose libgeos-dev
 ```
 
-## Install optional dependencies to be able to use nrml validator from risklib package
+## Change the PostgreSQL configuration
+
```
micheles: PostgreSQL? Is this documentation for old versions of the engine?

daniviga: Michele, we are talking about the Platform, which uses Postgres

docs/Development-environment-on-Ubuntu-12.04.md
```diff
@@ -6,32 +6,47 @@ These are the main steps to install the OpenQuake Platform on Ubuntu 12.04 LTS u
...
 pip install 'http://github.com/gem/oq-hazardlib/tarball/master'
 pip install 'http://github.com/gem/oq-engine/tarball/master'
 
-# OR download 'oq-hazardlib' and 'oq-engine' manually from github and make available via PYTHONPATH
-# before run any python applications
+```
 
-# ---- else if your virtual environment HASN'T ACCESS to system-wide packages ----
+#### Using a virtualenv without access to the global site-packages
```
nastasi-oq: "the global" => "the system-wide" maybe is more clear.

docs/Development-environment-on-Ubuntu-12.04.md
```diff
@@ -6,81 +6,78 @@ These are the main steps to install the OpenQuake Platform on Ubuntu 12.04 LTS u
...
 
+```bash
 # you can install all required packages using pip:
 pip install futures
 pip install decorator
 pip install psutil
 pip install Cython
 pip install h5py
-pip install selenium
```
daniviga: @nastasi-oq I removed selenium from here because it is not an engine dep. But maybe we should add a section about installing everything is needed to run selenium.

docs/Development-environment-on-Ubuntu-12.04.md
daniviga: This comes from the `virtualenv` help

```
  --no-site-packages    DEPRECATED. Retained only for backward compatibility.
                        Not having access to global site-packages is now the
                        default behavior.
  --system-site-packages
                        Give the virtual environment access to the global
                        site-packages.
```

hascar:
LGTM

nastasi-oq:
LGTM

PR #621, opened 2016-11-23
IRV: enable reading projects using a modified custom operator format

ptormene:
I am keeping backwards compatibility, so the old projects can still be visualized.
With the proposed changes, the IRV will be able to read also projects created by the IRMT QGIS plugin after the changes introduced in the following commit:
https://github.com/gem/oq-irmt-qgis/commit/03b8a3d15c31cdc7b852a7b3ad4aa0bdade8a6e0
While displaying the tree, the IRV will visualize also the custom formula or its textual description, if they are specified in the project definition.
Jasmine tests are green.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
-python-oq-platform (1.8.0)
+python-oq-platform (1.7.2)
   [Paolo Tormene]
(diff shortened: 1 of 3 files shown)


daniviga:
LGTM

PR #640, opened 2017-04-02
Fix tax base

nastasi-oq:
A new parametric variable for 'taxtweb' must be created in 'vulnerability' too, because it use taxtweb JS API.
Tests are green here: https://ci.openquake.org/job/zdevel_oq-platform/414/


hascar:
LGTM

PR #648, opened 2017-11-05
Add python3compat.py

daniviga:
Required by https://github.com/gem/oq-platform-ipt/pull/35


micheles:
So we have to add this twice?

PR #650, opened 2017-11-17
vuln: add taxonomy also when is not GEM type

nastasi-oq:
Add taxonomy to NRML also when taxonomy type is not GEM.
Tests are unavailable currently because tests infrastructure is not yet aligned.


hascar (approved):
LGTM

PR #651, opened 2017-11-17
vulnerabilty:fuln-function with 'vulnerabilityFunctionID' with taxonomy

nastasi-oq:
vulnerabilityFunctionID must be set with taxonomy


daniviga (approved):
LGTM

PR #654, opened 2017-11-29
Apps versions

nastasi-oq:
Add new unlinked /versions/ page with the list of our public applications and their versions.

### DIFF
--- a/openquakeplatform/openquakeplatform/common/templates/common/versions.html
+++ b/openquakeplatform/openquakeplatform/common/templates/common/versions.html
@@ -42,8 +42,8 @@
   <h2 class="page-title">{{page_title}}</h2>
   <div class="versions_subcontent">
     <table class="versions_table">
-      {% for item in apps_list %}
-      <tr><td>{{ item.name }}</td><td class="second">{{ item.vers }}</td></tr>
+      {% for app in apps_list %}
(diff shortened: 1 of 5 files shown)


openquakeplatform/openquakeplatform/common/views.py
```diff
@@ -0,0 +1,37 @@
...
+        app = import_module(app_name)
+        header_info = getattr(app, 'header_info', None)
+        vers = getattr(app, '__version__', None)
+
+        if header_info:
+            name = header_info['title']
+        else:
+            name = app_name
+
+        if vers is None:
```
ptormene: Perhaps this check should be moved before the previous one? If you don't do the append, you don't need to get the name.

openquakeplatform/openquakeplatform/common/templates/common/versions.html
```diff
@@ -0,0 +1,51 @@
...
+</style>
+{{ block.super }}
+{% endblock %}
+
+{% block middle %}
+<div class="versions_content">
+  <h2 class="page-title">{{page_title}}</h2>
+  <div class="versions_subcontent">
+    <table class="versions_table">
+      {% for item in apps_list %}
```
ptormene: It's fine as it is, but perhaps you could rename item -> app, just for readability

ptormene (approved):
Apart from a couple of very minor comments, LGTM

